PR #53, opened 2016-09-15
change return value of authenticate; add id

nicolasfranck:
Changes:

* return value of LibreCat::Auth: hash or undef. So it is still the same in a boolean context.
* return value of LibreCat::Auth returns information about the module and the user:

```
{
    uid => "njfranck",
    package => "LibreCat::Auth::Bag",
    package_id => "LibreCat::Auth::Bag"
}
```

* add method "id" to LibreCat::Auth. Defaults to the package name. This is usefull when using
  multiple instances of the same package in LibreCat::Auth::Multi, and the application
  needs to know which package succeeded. This is "id" is stored in the key "package_id" of the returned hash.
* add method "id" to LibreCat::Auth::SSO
* LibreCat::Auth::SSO now stores similar hash in session as LibreCat::Auth (with the addition of "response")

### DIFF
--- a/lib/LibreCat/Auth.pm
+++ b/lib/LibreCat/Auth.pm
@@ -9,6 +9,12 @@ with 'Catmandu::Logger';
 requires '_authenticate';
 
 has obfuscate_params => (is => 'lazy');
+has id => (
+    is => 'ro',
+    lazy => 1,
+    required => 1,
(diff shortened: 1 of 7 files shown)


vpeil:
looks good to me. thanks  @nicolasfranck

Issue #151, opened 2017-02-02
Return button in the edit form doens't update the publication


phochste:
Reviewers complain in Ghent that publication updates don't get saved (e.g. comments) when clicking on the "return" button.

Also they would like to see records automatically unlocked when the "return" is clicked.

vpeil:
The return button does only this: changes the status from public to return. No other fields are updated.
I am open to change this behviour.

vpeil:
fixed with PR #173

PR #173, opened 2017-02-21
save form values when clicking returning a record

nics:
Up to now form values are discard when returning a record. this pr makes return behave the same as save or make public. Form values are saved.


vpeil:
this PR solves #151

Issue #200, opened 2017-03-13
CLI Crossref mapping


njahn82:
The crossref importer is not mapping journal titles (`publication`). Maybe the fix file needs a fix.

Thanks!

Reprex:

```
bin/librecat publication fetch crossref 10.1016/j.conb.2016.12.004
```

njahn82:
Here's the API output:

```
{
    "status": "ok",
    "message-type": "work",
    "message-version": "1.0.0",
    "message": {
        "indexed": {
            "date-parts": [
                [
                    2017,
                    2,
                    16
                ]
            ],
            "date-time": "2017-02-16T08:40:19Z",
            "timestamp": 1487234419501
        },
        "reference-count": 81,
        "publisher": "Elsevier BV",
        "license": [
            {
                "URL": "http://www.elsevier.com/tdm/userlicense/1.0/",
                "start": {
                    "date-parts": [
                        [
                            2017,
                            2,
                            1
                        ]
                    ],
                    "date-time": "2017-02-01T00:00:00Z",
                    "timestamp": 1485907200000
                },
                "delay-in-days": 0,
                "content-version": "tdm"
            }
        ],
        "content-domain": {
            "domain": [
                "clinicalkey.jp",
                "clinicalkey.com",
                "clinicalkey.es",
                "clinicalkey.com.au",
                "clinicalkey.fr",
                "bmn.com",
                "elsevier.com",
                "sciencedirect.com"
            ],
            "crossmark-restriction": false
        },
        "short-container-title": [
            "Current Opinion in Neurobiology"
        ],
        "cited-count": 0,
        "published-print": {
            "date-parts": [
                [
                    2017,
                    2
                ]
            ]
        },
        "DOI": "10.1016/j.conb.2016.12.004",
        "type": "journal-article",
        "created": {
            "date-parts": [
                [
                    2016,
                    12,
                    26
                ]
            ],
            "date-time": "2016-12-26T20:30:26Z",
            "timestamp": 1482784226000
        },
        "page": "102-110",
        "update-policy": "http://dx.doi.org/10.1016/elsevier_cm_policy",
        "source": "CrossRef",
        "title": [
            "Synapse biology in the ‘circuit-age’—paths toward molecular connectomics"
        ],
        "prefix": "http://id.crossref.org/prefix/10.1016",
        "volume": "42",
        "author": [
            {
                "given": "Dietmar",
                "family": "Schreiner",
                "affiliation": []
            },
            {
[...]

PR #204, opened 2017-03-14
Update crossref_mapping.fix

christianmahnke:
Change the `-` to `_` since this is hard coded in line 9. Might help to solve #200. Now the `publication` field is populated correctly.


vpeil:
ah yes, I guess this one fixes #200

PR #205, opened 2017-03-15
Update crossref_mapping.fix

christianmahnke:
This change fixes an issue, that is documented [here](https://github.com/subugoe/goefis/issues/115). This is a typing error (integer instead of string), we 'reuse' `capitalize()` as the Catmandu documentation doesn't show a easy way to do type casts (it seems there is even a concept of data typing missing).

### DIFF
--- a/fixes/crossref_mapping.fix
+++ b/fixes/crossref_mapping.fix
@@ -71,3 +71,6 @@ remove_field(issued)
 remove_field(member)
 remove_field(container-title)
 remove_field(funder)
+if is_number(year)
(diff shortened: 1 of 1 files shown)


vpeil:
you force the year field to string?

njahn82:
Our templates are breaking after import. How can we convert it to a
datetime object?

njahn82:
Forget it, Patrick has explained to us why our import crashed:
https://github.com/subugoe/goefis/issues/115#issuecomment-286978980

PR #206, opened 2017-03-16
Extend locale.en.yml

dbeucke:


vpeil:
there's a typo in line 1225 lable -->label

PR #218, opened 2017-03-19
Update features.yml

christianmahnke:
Also use facets in the frontend. This change makes the off the shelf LibreCat a bit more like https://pub.uni-bielefeld.de/publication

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -7,7 +7,15 @@ views:
   frontend:
     publication:
       facets:
-        - - template: facets/popular_science
+        - - template: facets/type
+            facet: type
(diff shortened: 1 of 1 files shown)


vpeil:
I think it is easier to put it in your layer, rather than having others removing it from their layers....
But, to do others think?

christianmahnke:
Well, it is actually a feature that you might want to showcase on demo.librecat.org.

vpeil:
yeah, I see the case for the demo server

vpeil:
I'll put it into a layer for the demo server

PR #240, opened 2017-03-27
Update cpanfile

christianmahnke:
See subugoe/goefis#138 for a description, why this change is needed.


christianmahnke:
I've also created a bug report for Catmandu on this: https://github.com/LibreCat/Catmandu/issues/299

PR #266, opened 2017-04-11
Dev

njahn82:
crossref importer defaulted every import to journal_article. We will have to evaluate how many records are affected.

### DIFF
--- a/fixes/crossref_mapping.fix
+++ b/fixes/crossref_mapping.fix
@@ -6,7 +6,7 @@ move_field('message','')
 add_field('status','new')
 
 # set default type
-add_field('type','journal_article')
+# add_field('type','journal_article')
 
 #title
 #move_field('subtitle.0','title.1')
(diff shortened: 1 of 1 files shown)


vpeil:
but you need the line 9 for a default value. otherwise you only get book, book_chapter and undef.

njahn82:
and `journal_article`. 

So far, every record is mapped to `journal_article` before the fix applies type-specific mappings. Just try this DOI, representing a book chapter, on the demo server

10.1007/978-3-531-93139-5_10
http://api.crossref.org/works/10.1007/978-3-531-93139-5_10

I'd rather prefer to have an undef because it indicates me that something went wrong than a default mapping to one type. We'll have to check thousands of records again where we simply collected the DOIs and used the LibreCat CLI for bulk import.

Will try to create a lookup table of publication types between Crossref and LibreCat.

vpeil:
oh, haven't seen line https://github.com/njahn82/LibreCat/blob/8367ce45bfe09f0830da75f9300a30ff46cfeeea/fixes/crossref_mapping.fix#L52

PR #289, opened 2017-08-08
csl is optional

nicolasfranck:
The Citation Style Language is marked as optional, so the default setting should be "no".


nicolasfranck:
When pull request 
https://github.com/LibreCat/LibreCat/pull/290
is merged, I shall resubmit this pull request. 
Because that's the reason why it fails ;-)

PR #292, opened 2017-08-16
LibreCat::Auth::SSO should use middleware rebase

nicolasfranck:
The former implementation depended on the Plack middleware "ReverseProxy", and tried to figure out the full urls  from that information.

Now "uri_base" needs to be configured explicitely, like in the main application.


vpeil:
thanks @nicolasfranck

PR #295, opened 2017-08-16
Make Dancer configuration overridable

nicolasfranck:
These files:
```
environments/*.yml
```
should not be included by default.

They should be used to make local changes to config.yml.


vpeil:
Agreed!

PR #296, opened 2017-08-17
Move configuration of route /export

nicolasfranck:
* Moved configuration of route /export from
```
Catmandu->config->{exporter}->{publication};
```
to
```
Catmandu->config->{route}->{exporter}->{publication};
```
I could have made:
```
Catmandu->config->{route}->{exporter}
```
but I'm keeping the options open to distinguish between different kinds of "bags"
(which is now "publication") in the future

* Created /publication/:id.:fmt route in publication.pm

   so removed need for "referer" in export.pm

* Created /data/:id.:fmt route in publication.pm
* Created /marked.:fmt in mark.pm

   Apparently /marked returns ALL marked records, in order to reorder them,
   while /marked.:fmt does not. Is this intended?


vpeil:
>>Apparently /marked returns ALL marked records, in order to reorder them,
while /marked.:fmt does not. Is this intended?<<

No, seems to be a bug here. I checked your PR: mark 2 publications, go to marked page and export there. What you see are more than the 2 publications (could be that the bug was already there before)

nicolasfranck:
@vpeil : the export links do not work anymore as they used to

In the past
```
/export?fmt=yaml
```
was seen as a request from the page "/marked" because it took the referrer into account.
But that's not the case anymore (too transparent), so it exports everything

I should use the new route "/marked?fmt=yaml"

vpeil:
thanks for clarifying! I'll change that in the export facets

nicolasfranck:
@vpeil : mm, I see that filters.tt is reused on multiple locations:
* home.tt
* publication/list.tt
* embed/iframe.tt
* marked/list.tt

create something else?

vpeil:
yes, creating its own marked/filters.tt is here the way to go

vpeil:
The route is called `/marked.:fmt`, but that's fine for me. changed the template.

Issue #124, opened 2017-01-24
Remove inline (CSS) styles


nilswindisch:
https://github.com/LibreCat/LibreCat/search?utf8=%E2%9C%93&q=style+in%3Afile+extension%3Att&type=Code

Styling is best done via CSS class names and related code in CSS files. Removing inline styling makes it easier to adapt LibreCat for institution specific look & feel.

vpeil:
thanks @verheyenkoen

PR #310, opened 2017-09-22
Feature remove inline css

verheyenkoen:
Closes issue #124


vpeil:
@verheyenkoen I really like what you did here.

In `scss/main.scss` you import bootstrap-4. Is that already used or is that for future development?

verheyenkoen:
@vpeil Bootstrap 4 is now in beta but it is already the version they promote from their front page. I added this file as a "polyfill" for future features. Once we upgrade to Bootstrap 4, that file can be removed again.

PR #312, opened 2017-10-02
Feature/seamless index switch

vpeil:


vpeil:
@phochste I think the idea was the following:
start with a fresh elasticsearch (`curl -XDELETE localhost:9200/*`)
then run `bin/librecat index initialize`.

phochste:
@vpeil The problem is when you execute `bin/librecat index initialize` a second time, the command crashes and both indexes are deleted. At least the command should work (re initialize something) or prevent you from initializing an index that is already ok.

vpeil:
@phochste please, have a look at the latest commit f7797a2

phochste:
This branch is too big to merge. I've created a new branch only containing the index switch code

PR #314, opened 2017-10-05
Pr/default hooks

vpeil:
These are the default_hooks introduced by @nics.

I have changed the order of execution as follows:

1. before_fixes
2. default_before_fixes
--> here happens the action of the fix_around sub
3. default_after_fixes
4. after_fixes

Tests for the code added + pod.


vpeil:
Indeed. Please merge #318 first.

phochste:
After discussion with Nicolas this pull request can be closed for now

PR #318, opened 2017-10-06
Bug upload file 317

vpeil:


phochste:
@Vitali This is not only about file uploads. There are more things in Catmandu::Plugin::LibreCat::PublicationStore that should be processed before the validation happens.

phochste:
Everything in   around add  , set_delete_status .. basically needs to happen before the hooks can get into action

Issue #219, opened 2017-03-20
RTF export is empty


dbeucke:

vpeil:
that is due to our change with the default citation style (Title + "short")....

nicolasfranck:
It's still empty in the current dev branch.

I noticed a unused method in Catmandu::Exporter::RTF: _add_cite.
That method is never called because it's overridden by the method with
the same name of Catmandu::Exporter::Cite ( put some log statements
inside these methods too see this!).

Reason: "extends" is called AFTER all perl methods are compiled, so _add_cite of the
child class is seen first, then _add_cite of the parent class.
Methods should be overridden with "around", so Moo can follow.

vpeil:
fixed by PR #325

PR #325, opened 2017-10-13
refactor RTF export; fixes GH #219

vpeil:
This fixes the bug in #219.

Things to consider for the near future:

- RTF export should be displayed in case of `Catmandu->config->{citation}->{engine} eq 'cls'` only. Otherwise there's nothing to export. (feature flag)


vpeil:
also removed hard-coded "bielefeld" links

PR #329, opened 2017-10-18
Pr/remove local names

vpeil:
still PUB/Bielefeld stuff found in code


verheyenkoen:
@phochste tells me these are also Bielefeld related names:
https://github.com/LibreCat/LibreCat/blob/dev/config/locale.en.yml#L196
https://github.com/LibreCat/LibreCat/blob/dev/config/locale.de.yml#L198

scss/print.scss
```diff
@@ -23,7 +23,7 @@
     }
 
     a[href^="/"]::after {
-        content: " (https://pub.uni-bielefeld.de" attr(href) ") ";
+        content: " (" attr(href) ") ";
```
verheyenkoen: I'm not sure this rule should remain. It will now print the following if the domain name is not in the href attribute:
image

Not sure if relative paths are helpful.

vpeil:
Yes. you are right, I should clean up the `config/locale.*.yml`

Please, fill free to change or remove the rule in scss/print.scss, since you are our frontend expert!

Issue #328, opened 2017-10-17
Missing quick link to marked publications on publication overview page


phochste:
I need to click always the details of one publication to find a list to all marked publications. Put also this link on the publication overview page.

verheyenkoen:
Where do you want this in the sidebar? 
![image](https://user-images.githubusercontent.com/599538/31828049-889f82fc-b5b9-11e7-92d2-6a05ca6155aa.png)

phochste:
For now under Export/Embed. This is the position the "Marked publication" has on the detail page.

PR #332, opened 2017-10-23
Added marked publications quick link to publications page (issue #328)

verheyenkoen:
addresses #328


phochste:
The location seems right for now. 

The details page seems to miss a new "back to result" list button. When I check "Mark/unmark" publication and return to the result list (browser history(-1)) of course the "Marked Publication(s)" on that page is not updated. Or, are there better ways to solve this (or..doesnt matter: its a feature)?

Issue #233, opened 2017-03-23
Cache author list (like departments)


vpeil:
The route `/person` is too slow.

vpeil:
The generation of static html for this purpose was a bad idea (see closed PR #340).

We should add another indicator (e.g. show=1) to the user index. This indicator can be calculated over night (show=1 if and only if person is related to "public" publications).

PR #340, opened 2017-11-06
Author lists 233

vpeil:
This PR solves #233 

Changes:
- Introduce new command `bin/librecat generate authors`


phochste:
Thanks for this solution. We have a discussion here in Ghent about the need of more generating templates in LibreCat. We do this now for departments and forms. There is a general mood that template generators are not something we like to keep in a next major release. 
I don't mind a temporary solutions, but we need to find out why these things have to be generated. Can there be other (future) solutions.

@nics mentions the possibility to have a cron/gearman job update a store table with publication statistics for users/departments on which the authors list and department list can be based. Preferably in a seperate store from departments and users (not to get into issues of multiple threads editing the same records constantly).

I don't know if there are other options.

vpeil:
Agreed, I would consider this a temporary solution.

I have  to confess that I'm not that happy with **all** of the template generation (e.g. supporting other insitutions makes it much harder, because one always has to keep this extra step in mind).

So, :+1: for a future change for department/author templates. And :+1: ^2 for dynamically generating the edit forms.

Please, decide if you would like to merge or close it.

vpeil:
I'd like to close this (I have another solution in mind).

PR #345, opened 2017-11-09
Issue 337

nicolasfranck:
What I changed:

* method signature:
```
  can_edit( $pub_id, { role => $role, user_id => $user_id } )

```

Should be done in the future:

* Split packages that combine packages. 

e.g. 
LibreCat::App::Catalogue::Controller::Permission
LibreCat::App::Helper

LibreCat::App::Catalogue::Controller::Permission needs Dancer
LibreCat::App::Catalogue::Controller::Permission::Permission doesn't need Dancer but gets it indirectly by using LibreCat::App::Helper
LibreCat::App::Helper::Helpers doesn't need Dancer


phochste:
Author checks are on `_id` field instead of the `id` field

phochste:
Thanks

PR #353, opened 2017-11-27
remove bielefeld background reference, images have been deleted earlier

vpeil:
what is left: need to re-generate main.css, (I am getting some npm errors on my installation)


vpeil:
@verheyenkoen The images files were in the repo (we removed it at some point). So my thought was also to cleanup up these references. Either we rename them, if you'd like to keep a background image or we remove this completely. 

As decribed, of course the main.css needs to be regenarted, but I wasn't able to do this.

verheyenkoen:
@vpeil I regenerated the css file for you.

vpeil:
thanks

phochste:
Thanks!

Issue #358, opened 2017-11-30
main_file_link is hard coded in the templates as a one item array (main_file_link.0)


phochste:
If main_file_link should be an array, there should also be support in the templates for multiple array values.

petrakohorst:
It is configurable. In the forms.yml you can set "multiple: 1" for the field "main_file_link" (here https://github.com/LibreCat/LibreCat/blob/master/config/forms.yml#L238 ) and re-generate the forms.
For the core it is currently set to be "multiple: 0".

phochste:
But as I understand only one main_file_link will be displayed in record.tt? If there is support for multiple main_file_link's , then they should end up in all displays?

vpeil:
@petrakohorst I guess, the forms configuration option "multiple: 0|1" has effect on edit forms only, right?
Not on public views, like frontdoor.

petrakohorst:
That is indeed true. Only the first link is shown on the frontdoor (record.tt). I never noticed that.

It is more of a design issue, though. I'll think of something.

vpeil:
fixed by PR #360

PR #360, opened 2017-12-01
Feature/display all main file links

petrakohorst:
Display all main_file_links on frontdoor (record.tt).
Links are now all displayed on tab file_details below any fulltext download links. In the upper section of the frontdoor still only the first url is dispayed.


phochste:
Thanks!

vpeil:
fixes #358

PR #372, opened 2018-01-05
correct schema entry status wihtin related record

vpeil:


vpeil:
oh wait, I messed something up here.

PR #375, opened 2018-01-10
Make cc licenses mandatory (never, always or for oa-files only) by config file.

petrakohorst:
Needs "bin/librecat generate forms".

The config.yml can now specify "mandatory: oa_file", to make cc licenses mandatory only if an oa file is present. Uploading a file (which is per default oa) will turn cc license field into a mandatory field that needs to have a value selected to be able to save the record.

Remaining bug: If an uploaded oa file (for which the cc license field was made mandatory) is changed to local or restricted access, the cc license field will remain mandatory. Saving the rec (with a selected license) and then editing the rec enables the user to remove the cc license again.


vpeil:
I need to figure out what
> The current dev branch has the problem that the mandatory flag has no effect

means:

- If you upload a new file (which will be set OA by default) license is mandatory. And this works as expected.

But we have some edge cases here:
- The first one @petrakohorst has described above.
- But the reverse is also true: If you have an existing record with an existing closed file, open it for editing, change the access_level to OA, **but** license will no be mandatory.
- In case of an embargo ... well that's getting really complicated, depends on what access level is set after embargo.

All of these demonstrate (again), that such things would be much easier to handle on a server side validation.

vpeil:
I must confess, that I asked Petra to implement this. But I haven't had these issues in mind. Please blame me for this :-1:

phochste:
@vpeil What I mean is that the current 'dev' branch the mandatory flag on the 'cc_license' has no effect. You can fill out whatever you don't see any difference.

The pull requests fixes this bug..but it works a bit too good in case of oa_file: it also asks cc licenses if all the files are not open access.

vpeil:
@phochste We should close this PR. Talked to  @petrakohorst ... better have license information on file level than on record level.

PR #376, opened 2018-01-11
fixes #363, can remove departments safely

### LINKED ISSUE #363: Adding/Removing of Department Affiliation fails
Several errors appear while adding/removing of department affliations at different pages.

To reproduce the error: I logged in with einstein at http://demo.librecat.org and changed the role to User.

- At http://demo.librecat.org/librecat I cannot remove the affliation (Department of Physics -> Relativity Theory Group) by clicking on Settings and then the minus: `An internal error occured`
- At http://demo.librecat.org/ and at http://demo.librecat.org/person/1234 Both, adding and removing affiliation via +/- fails with a JavaScript error:
`Uncaught ReferenceError: add_field is not defined at HTMLDivElement.onclick (1234:118)`
`Uncaught ReferenceError: remove_field is not defined at HTMLDivElement.onclick (1234:119)`

vpeil:
This PR fixes #363 

The only thing is: I can't see any JS-errors (also before the PR!).


feldi88:
@vpeil According to the JS-errors:

I looked into it again: It really depends on which view you are.
The errors only appear if you are at the url http://demo.librecat.org/ and not if you are at the url http://demo.librecat.org/librecat I also investigated deeper and found the problem: Depending on the view, different javascript files are loaded: https://github.com/LibreCat/LibreCat/blob/8704297793ca545d7cae185a9a36201d5e1ce182/views/head_meta.tt#L20

At the working views, https://github.com/LibreCat/LibreCat/blob/master/public/javascripts/backend.js is loaded while at the non-working views, https://github.com/LibreCat/LibreCat/blob/master/public/javascripts/frontend.js is loaded instead. While the first one has defined the JavaScript method `add_field`, the second one hasn't. 

Perhaps methods like this should be extracted to a common.js file or at least the backend.js has to be loaded if a user is logged in at all views and not only at some.

vpeil:
Ok, now I got it! Thanks for investigating.

vpeil:
@verheyenkoen need your opinion:
backend.js is 34K small, frontend.js is 1.8K small. Instead of introducing a third file for this purpose (see suggestion above), wouldn't it be more appropriate to put these into one file. Besides fixing this error template will be cleaner as well.

vpeil:
@phochste the latest commits on this branch solve the js-errors. Any objections from your point of view?

phochste:
Thanks

PR #393, opened 2018-02-07
change host to uri_base

vpeil:
make sure that sitemap urls have absolute paths


phochste:
Thanks++ (I can keep repeating just imagine it is there ;)

PR #406, opened 2018-03-07
see #233, prepare user search: makes retrieval much faster

vpeil:
this solves #233 (but not like described in the issue).

Please provide some feedback.


vpeil:
Do we want to display publication counts? In that case I have to modify this PR a bit.

phochste:
Yes, the publication count is shown in the departments tab. It would preferably also be present in the authors and projects tab.

vpeil:
@phochste requested changes implemented.

phochste:
Ok thanks
@verheyenkoen Can you update the cypress tests? In new `dev` branch the test will fail to dynamically load the user count on the Manage User Accounts page and User tab on the Front page

verheyenkoen:
@phochste Removed the specific test. Was testing for the AJAX to work which is now removed so no use in keeping a test for that.

PR #407, opened 2018-03-08
Pr/layer friendly templates

petrakohorst:
Some small changes to make templates easier to handle in layers.

E.g.
* Moving some template parts to smaller subtemplates so they can be replaced easily in a layer (without having to replace the whole parent template).
* Moving key "appname_short" to language files so it can be overwritten in layers (appname_short is the name of the link to the frontdoor in publication list)


vpeil:
I would sugget to remove appname_short from config.yml. But then one needs to cleanup some templates:
run `grep -r appname_short`.

petrakohorst:
I'll check if I can remove appname_short from config.yml and set the value from the language files instead.

vpeil:
@phochste all check marks are green. Please, have a look

Issue #420, opened 2018-03-16
JavaScript embed produces invalid JavaScript


feldi88:
The JavaScript embedding option produces invalid JavaScript. It contains the citation data in a `document.write`, but with an invalid line break. For example, the following is produced:

```
<!-- BEGIN embed/javascript.tt -->
document.write ('<ul>') ;
document.write('<li>');
document.write('Krijger, Stefanie, <em>Listening challenges for children with cochlear implants in mainstream education</em>. Royal Belgian Society for Ear, Nose, Throat, Head and Neck Surgery, Spring meeting, Abstracts <strong></strong> (). , 2016<br />test
');
document.write('</li>');
```

The linebreak between the fourth and fifth line is invalid, so the HTML will not be rendered in most browsers.

The problem is probably the template `INCLUDE ` in https://github.com/LibreCat/LibreCat/blob/2a54792d2c06a27874fc3b212fd3f87ab78a2f84/views/embed/javascript.tt#L14 and https://github.com/LibreCat/LibreCat/blob/2a54792d2c06a27874fc3b212fd3f87ab78a2f84/views/embed/javascript.tt#L16 which produces and extra linebreak in the output.

vpeil:
solved with PR #421

PR #421, opened 2018-03-16
Issue 420

vpeil:
solves #420 

and I have added a test to avoid this in the future


phochste:
This `views/embed/links_js.tt` is pretty much unreadable/maintainable in its current form. Template toolkit allows to use '[%-' and '-%]' tag delimeters to remover trainling whitespace. Can you check how this code could be rewritten. Many editors just add newlines by opening the file. It would be bad to be dependent on this behavior.

vpeil:
@phochste I have pushed a solution as suggested.

PR #422, opened 2018-03-21
Make parameters in helper function 'extract_params' configurable

petrakohorst:
Sometimes you need additional parameters in a layer and you want them to be extracted by the helper function 'extract_params'.

I've changed this helper function to read params from a config file instead of listing them in the function itself. Also I've used a hash in the config, not an array, so layers don't have to overwrite the whole thing.

Each layer can add parameters in the key 'extract_params' in a config file called helper.yml. Those parameters will then be extracted as well.


vpeil:
I think I made the requested changes

PR #434, opened 2018-04-03
Improve BibTeX import

michaellass:
* Remove all curly braces from title. Example title that until now caused issues: "{ATUAE}: A Title Used As Example"
* Use data from publisher and school fields to determine publisher
* Properly import volume
* Translate -- to - in page range

This overall brings the import also much closer to the export logic defined in to_bibtex.fix.

I could not test the changes myself since I am only a user of LibreCat and have no local installation at hand.


phochste:
Thanks for the pull request

PR #436, opened 2018-04-06
Pr/rm data route

vpeil:
I am afraid that this PR is a little too big, but it tackles to desired change as a whole.

The data route is removed, the route publication is replaced by record. Both are left as redirects for backwards compatibility.

If this PR is approved we need to modify the cypress frontend tests.


vpeil:
@phochste removed parts as requested. Thanks for your open eyes!

vpeil:
conflicts resolve with https://github.com/LibreCat/LibreCat/pull/436/commits/0ec2d1c9ae4621e867cca841b38ae969382a5f8f

PR #445, opened 2018-04-09
Refactor CRUD functions into model objects

nics:
This pr refactors the CRUD operations in App/Helper.pm into model objects, greatly cleaning up the code.
LibreCat.pm is now a proper object that holds references to these model instances.
The functionality of Layers.pm is now also contained in LibreCat.pm, making it the central object.


phochste:
thx!

PR #466, opened 2018-04-26
Pr/issue 418 2nd edition

### LINKED ISSUE #418: Publications with only open access main_file_links not being recognized as open access
I have publications with only main_file_links and no file uploaded locally. Open_access=1 in these publications but they are not being recognized as open access in  the interface. the 'Open Access Publications' link says 0.

vpeil:
This one fixes #418. Re-indexing is required here


phochste:
@vpeil Can you check the conflicts in this branch?

vpeil:
resolved merge conflict.

phochste:
Checking the code currently. The oa:1 only currently ends up in the EL index when all records get reindexed, but now when a single record is edited via the webforms

vpeil:
> Checking the code currently. The oa:1 only currently ends up in the EL index when all records get reindexed, but now when a single record is edited via the webforms

That's a strange behaviour. The `index_publication.fix` should be called on each add, right? Through the Model add method.

Issue #472, opened 2018-05-02
Renaming of projects causes inconsistent data


feldi-online:
I have an existing project with name "A" and tag a publication with this project and save it. Everything is fine.

Then I rename the project (for example fix a typo or upate the name in general), for example to "B". But I do not touch the publication. In the publication data the old project name is still used, so it is still "A". This inconsistency can only be resolved by removing the project from the publication and adding it again with the new name. But this is not a usable way.

So after the renaming of projects a reindexing/fix or something else has to be run to update all tagged publications.

PR #485, opened 2018-05-17
Fixing #472 project names in publication shouldt have any effect

phochste:
Projects in publication are only referenced by their id. In webpages/exports these ids are matched against the project database to fetch the current name. Exciting project names in the publications will have no effect anymore, they can be safely deleted.

### DIFF
--- a/config/schemas.yml
+++ b/config/schemas.yml
@@ -229,9 +229,6 @@ publication:
       type: object
       properties:
         _id: *id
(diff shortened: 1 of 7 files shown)


vpeil:
> Exciting project names

Probably, you mean "existing project names"

phochste:
@vpeil maybe both :)

Issue #284, opened 2017-06-09
DOI and ISBN registration


phochste:
Ref. Wageningen's services http://elag2017.org/?option=com_sppagebuilder&view=page&id=46

vpeil:
DOI registration is supported via `LibreCat::Worker::DataCite` and the hook `LibreCat::Hook::new_research_data_datacite` (maybe choose a better name, sth. `...register`), but has been applied to type=research_data only.

We could hook that in whereever we want (also in layer!), or think about a checkbox in edit forms 
like ("Give me a DOI :white_check_mark: )

ISBN registry looks weird. Can't see any API or documenation and it depends on the country???
e.g. http://www.german-isbn.de/

vpeil:
A solution for the DOI registration is available on branch [doi_register](https://github.com/LibreCat/LibreCat/tree/pr/doi_register)
I think this branch needs a bit of testing, then this will be suggested as a PR.

I have written a small documentation: https://github.com/LibreCat/LibreCat/wiki/DOI-Registration

Please let me know, if something is missing.

patrickda:
For the ISBN an institution can by a package with a distinguished prefix and i think the llast number has to be calculated. 
We would be also interested in the feature for our thesis. 
Give me a few days and I will provide more information about the ISBN.

vpeil:
@patrickda you mean that **one does not have to** push data anywhere? That would make things very easy.
Yes, the last digit is a checksum, all you need as provided by the perl module [Business::ISBN](https://metacpan.org/pod/Business::ISBN)

patrickda:
As far as i know,  this is not necessary. We got a block of ISBNs (the block is defined by the third number sequence, the second is the country. the fourth block can be just a number which is counting up. Therefor it just must be possible anywhere to define the registered block, so that a new ISBN can be created from the block ([see also Wikipedia](https://en.wikipedia.org/wiki/International_Standard_Book_Number#How_ISBNs_are_issued)) also an automated information if the block runs out would be cool. At last we would only use this feature for a specific item type (thesis).  

Still the whole item (book /ebook) has to send to the national library for documentation.

vpeil:
Sending stuff to a national library is up to local implementations, I guess. You would do this via `LibreCat::Hook`s.

vpeil:
I think this is done since the merged of PR #493 .

For ISBNs, either see discussion above or open a new ticket. But then, one has to provide some information.

PR #493, opened 2018-06-05
Pr/doi register

vpeil:
Solve #284 partially (DOI regsitration)


vpeil:
@phochste changes made as requested

vpeil:
@phochste this on is ready to merge (and the longer it is here, the more work I have, to keep that one in sync with dev branch).

phochste:
I need some feedback from @vpeil or @nics It was for me quite hard to test and debug this code.

In https://github.com/LibreCat/LibreCat/blob/pr/doi_register/lib/LibreCat/Hook/register_doi.pm#L23 a `librecat->log->debug` is used which results in no logging output when setting the log4perl parameters. I don't know why this is.

For @vpeil :

-  I assume one needs to add a `register_doi` hook added in `hooks.yml` `hooks.publication-publish.after_fixes` to make this work? (Or what is the best place?)
-  Also the `forms.yml` needs to be edited to activate the registration for `doi` fields
- The `LibreCat::Hook::register_doi` requires a seperate `hook.register_doi` configuration which is not defined and the supplied example doesn't work with the standard DOI prefix defined in `catmandu.local.yml-example`

..This `hook.register_doi` is actually what seems to be the `catmandu.local.yml-example` `doi` configuration. Can't this YAML section not be (re)used?

vpeil:
About the logging: no idea why this happens.

In #284 I have mentioned the wiki: https://github.com/LibreCat/LibreCat/wiki/DOI-Registration. Thjis article should contain everything one needs. Should it also be documented somewhere in the code?

I will have a look at the example configuration.

vpeil:
@phochste this commit provides the suggested config modification
https://github.com/LibreCat/LibreCat/pull/493/commits/c7da26b3b7b889ef90645a0ea5ff65b3ef7a71ea

phochste:
@vpeil About the logging: @nics  helped me:

This code only works when the top level `LibreCat` is set to `DEBUG` mode:

```
librecat->log->debug("...");  
```

This code will work when the `LibreCat::Pkg::Whatever` is set to `DEBUG` mode:

```
with LiberCat::Logger

sub  ... {
   
   $self->log->debug("...")
}
```

PR #504, opened 2018-06-11
Pr/custom mount path

vpeil:


nicolasfranck:
Wouldn't it be better to do this in the proxy server? I.e. set the prefix there?

nicolasfranck:
With only "/whatever", some paths will not be caught by Dancer, but by Plack itself. So your "page not found" would be different

vpeil:
I had this as a feature request from an institution using a load balancer. Of course, one needs to set the `uri_base` accordingly. One can do load balancer -> apache reverse proxy -> librecat web service, but the extra layer with apache reverse proxy seems to be superfluous.

nicolasfranck:
You can also make port 5001 available for the load balancer only (using firewall settings), and set "uri_base" (that's what it is made for) in the config to match the prefix from the load balancer. If understand correctly, you're starting plack on port 80?

vpeil:
aha, looks like a better solution. We will try this.

vpeil:
Please, wait with this PR.

nicolasfranck:
@vpeil : what have you decided?

PR #508, opened 2018-06-14
Fixes #506 abstracts without selected language

### LINKED ISSUE #506: Abstract is not saved when no language is selected
- Fill in an abstract but don't set a language
- Save the record
- The abstract is not saved in the database

petrakohorst:
Abstracts without language selection were not stored.

After this PR: if no language is selected, the fix will assign English to the abstract. In a new rec, English will be preselected for the abstract, but removed if no abstract text is added.

The general language field of a record will also have English preselected, so that any record will have a language.


petrakohorst:
@nicolasfranck : "eng" is correct. Take a look at the lists.yml. We use the three-letter codes for the language fields. (But please don't ask me WHY we use those here. :wink: )

PR #517, opened 2018-06-21
Set lang

nicolasfranck:
What it does:

* dancer-before-hook looks for parameter "lang", and stores it in cookie "lang"
* by storing "lang" in a separate cookie, the language survives between signin/signout.

Problems:

* when not set in the url, the language is implicitly determined by this cookie
* I tried using the history api to push this cookie as a parameter back in the url (for the copy-and-paste situation), but this has other problems ( messing up the history, leading to weird behaviour when going back ). See views/head_meta.tt

Maybe the language should be added to EVERY url, but there are so many urls. In that case no cookie would be required.


nicolasfranck:
or: the helper method "uri_for" can do this. It is already used everywhere

nicolasfranck:
What I've done:
* cookie "lang" is set when parameter "lang" is given in the url (happens in a dancer hook)
* helper.uri_for adds parameter "lang" to the list if configuration option i18n.show_locale is set to true (non existant means "true here)
* changes a few urls in the views
* so lang is always stored as cookie, but only shown when configured as such (edited)
TODO and questions:
* what about links in generated javascript files? Those use the global variable "librecat.uri_base"
* /set_language (NOT: the /set_language of one /person) is not used anymore. The language-link is the current link with the parameter "lang" set. How does Cypress test this? By clicking the url and check if the language has changed? (edited)
* some views do not have internationalisation (admin/account.tt, admin/project.tt, admin/research_group.tt)
Also some minor changes like html-encoding of values in the html-attributes (e.g. <a href="" title="..">)

nicolasfranck:
Will be split in more branches..

nicolasfranck:
Part 1: https://github.com/LibreCat/LibreCat/pull/524

PR #525, opened 2018-06-28
use content-disposition

nicolasfranck:
* use id-file.extension as suggested filename
* don't use the internal filename anymore
TODO: deprecate old route that uses the file_name

Issues:
* Chrome sometimes ignores the header content-disposition. Try uploading a file with extension ".txt", and download it.


nicolasfranck:
* content-disposition "filename" is configurable
```
files:
  download_file_name: "%i-%f.%e"
```

* %i: publication id
* %o: file_name
* %f: file_id
* %e: extension

Default is hardcoded as such:
```
"%o"
```
to be backward compatible

nicolasfranck:
@phochste : LibreCat::App::Helper::file_extension extracts the extension with the dot included. Is that supposed to happen?

PR #526, opened 2018-06-29
change type field to "password"

nicolasfranck:


nicolasfranck:
* create two routes:
  * PUT /librecat/admin/account/:id (UPDATE ROUTE)
  * POST /librecat/admin/account (CREATE ROUTE)
* remove route /librecat/admin/account/update
* generate _id in update route
* remove field "_id" from the form (both "edit" and "new"). Because this field is no longer present in the form "new", one can differentiate between both situation (instead or relying on the attribute "full_name" to be filled in..)

phochste:
Great. Very interesting this new Plack::Middleware::MethodOverride and easy methods to to PUT, DELETE , POST , PATCH and all that in the forms. This will be quite a crucial addition for later API developments in Librecat

PR #533, opened 2018-07-10
use h.loc instead of old lf.bla [was: show correct label for doc types]

vpeil:


vpeil:
@phochste the title of this PR was misleading. It is just a small change to use `h.loc` for lables instead of the old was `lf.bla.x.y.z`

You can merge safely.

Issue #530, opened 2018-07-04
External Identifiers in Links section of edit form


feldi-online:
The grouping of fields in the edit form is a little bit confusing, since the external identifiers are rendered in the "Additional Information" block, instead of in the "Links" block. I would prefer if they are put in the Links block because they are actually external links to other web services like the fulltext url or other external material. What do you think about this suggestion?

vpeil:
Sorry, I do not understand. Please, can you attach a screenshot with a marker?

feldi-online:
Yes, of course. The suggestion is to move the field to the last block (see screenshot):

![image](https://user-images.githubusercontent.com/6943048/42595091-68199168-8551-11e8-964a-db6eb32300af.png)

petrakohorst:
I have made the necessary changes to the external_id template, so that you can now configure this field to be in the links section (but only this field).

For more details, see PR.

PR #534, opened 2018-07-13
Allow external_id field to be placed in related_material section of the form.

petrakohorst:
The forms are meant to be flexible in placing fields in the different parts of the form. But so far this is only implemented for basic_fields and supplementary_fields. Fields can be moved between these two parts easily, but not the other two.

This PR solves #530 for the field external_id.

To place the field in the related_material section of a form, edit the forms.yml and move the definition of the field from ```publication_types.{type}.fields.supplementary_fields``` to ```publication_types.{type}.fields.related_material```. Then run ```bin/librecat generate forms```.

If more fields need to be moved from basic_fields or supplementary_fields to one of the other two sections, look at the diff in this PR and recreate what I have done for external_id.

### DIFF
--- a/config/forms.yml
+++ b/config/forms.yml
@@ -113,6 +113,7 @@ field_order:
     - related_material.record
     - related_material.rd
     - main_file_link
+    - external_id
     - related_material.link
 
 # file dropzone fields that may be changed/edited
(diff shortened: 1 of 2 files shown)


vpeil:
Basically, can't that be postponed until our rewrite of forms handling (for more detail see https://github.com/LibreCat/librecat-hfh-demo) ?

PR #537, opened 2018-07-25
belongs to #420; forgot to change this line

vpeil:


vpeil:
Have a look at https://github.com/LibreCat/LibreCat/pull/421/files and #420.

This just the same thing. Including templates in a js caused problems because of linebreaks. Now it implements this safely.

vpeil:
And the file  `embed/links_js.tt` is gone anyway since PR #421

PR #538, opened 2018-07-26
Fixes page numbers being lost when saving

petrakohorst:
When saving a record, neither the page(range), nor the article number were being saved.

Reason: the validator runs before the fix moves the values from the field called "page_range_number" to the fields "page" and "article_number". And the validator didn't know a field called "page_range_number", so it deleted the field.

### DIFF
--- a/config/schemas.yml
+++ b/config/schemas.yml
@@ -518,6 +518,8 @@ publication:
         - type: string
         - type: integer
       description: "Page number"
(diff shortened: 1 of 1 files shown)


phochste:
This is a bug in the order of hooks executed for every record, not a missing field in the schema. 
I'll close this pull request and add another fix for #566

PR #540, opened 2018-07-31
Remove html encodig from abstract on publication frontdoor

petrakohorst:
The abstract field should support markdown and html formatting.

With the ```| html``` encoding, however, html tags are being printed verbatim and not interpreted by the browser for formatting.


petrakohorst:
Closing this PR, changes will be done in local layer, as discussed in #547

Issue #500, opened 2018-06-07
"Search similar" should not change title to search


nicolasfranck:
In views/links.tt, the title to be searched is filtered:

```
<a href='[% uri_base %]/librecat/search/admin/similar_search?q="[% entry.title.replace("[\'\"\,\;\:\.]+","") %]"'>[% h.loc("main_page.links.search_similar") %]</a>
```

that's a job for ElasticSearch. This should be sufficient:

```
<a href='[% uri_base %]/librecat/search/admin/similar_search?q="[% entry.title | uri %]"'>[% h.loc("main_page.links.search_similar") %]</a>
```

I found this problem because I'm using file names as titles, and e.g. "input.jpeg" was changed to "inputjpeg", which could not be found.

See: https://github.com/LibreCat/LibreCat/blob/dev/views/links.tt#L68

vpeil:
Even if it would be necessary, it should not be contained in the templates.

nicolasfranck:
by the way, why is this route necessary? One could simply create a link to current search page with the parameter "q" filled. I see that this route similar_search creates a special query to match less accurate on titles, but shouldn't that be done in the main search page /admin/librecat (where there is also a filter on the deleted records!). Maybe set the accuracy on the field "title"?

Image someone just accessing the page "/admin/search", typing in a title, and expecting to find all titles that look like his/her query. Now you first need to find a specific record, and then this functionality is available.

petrakohorst:
```[% entry.title | uri %]``` looks good, but it is not sufficient. It doesn't solve the problem with the title "input.jpeg". Even with the uri filter, the title "input jpeg" is not found when searching for "input.jpeg". To solve this we should rather consider replacing punctuation with whitespace instead of nothing. But then only one direction will work. Searching for "input.jpeg" will find "input jpeg" but not the other way around. It remains a problem.

Also: without the replace and with only the uri filter, titles with single quotes will be cut off (due to the template quotes). So, at least single and double quotes need to be removed here.

Unfortunately this replacement needs to be done in the template, because of the single/double quote problem mentioned above. It does not suffice to replace or remove these characters later on in the route. Because the whole search term doesn't even make it to the route in the first place.

Example:

```Full publication 'journal article': the first one``` will be cut off in the template after the word publication and the following space.

As for the idea of using the parameter q and the regular search: in general I agree, that would be better. But I don't see how one would integrate the native search with match percentages into the regular search.

[...]

PR #541, opened 2018-07-31
Redirect to /librecat from similar_search if no search params

petrakohorst:
Redirect to /librecat from /similar_search if there are no search params. Otherwise the user will land on a similar_search page with no hits (e.g. after deleting the search terms).

### DIFF
--- a/lib/LibreCat/App/Catalogue/Route/search.pm
+++ b/lib/LibreCat/App/Catalogue/Route/search.pm
@@ -48,6 +48,7 @@ Performs search for similar titles, admin only
     get '/admin/similar_search' => sub {
 
         my $p = h->extract_params();
+        redirect '/librecat' unless keys %$p;
 
         # TODO filter out deleted recs
         my $hits = searcher->native_search(


petrakohorst:
Users will not know how to get back from similar search. 

They won't delete params from the URL, but they will try deleting the search term in the search box to get back to their normal view and their full list of publications.

With an empty "q=" the result is empty, that is correct. But then the user is sitting there with no publications to display, no filter to remove, no search term to remove. (Trust me, users will get stuck there! :wink: )

(@nicolasfranck alerted me to this issue in his last remark here: https://github.com/LibreCat/LibreCat/issues/500#issuecomment-413868888 )

nicolasfranck:
@petrakohorst : ah, I think you misunderstood my point ;-)

If you land on /similar_search, you don't know that the page has changed (you have to look at the url) because they look so similar (no pun intended). So if you change the query in the query-box, and hit enter, the form will be sent again to the controller /similar_search, because the form has no explicit "action" set. The behaviour in this page is a bit different from the regular page /publication, but the user has no clue why.

That is why I suggested to erase this route, and always create a link from "find similar title" to the same page "/publication" with the parameter "q" set. Maybe query a different analyzed field with more fuziness?

petrakohorst:
Yeah, ok, I see your point. The problem is not just an empty "q=", but the fact that any action on the page will keep the user in the similar_search universe.

In our local system we don't display the search term in the search box, but we show a delete link above the box to remove separate parts of the current query. So for our users that's the most common way back to the full publication list. That's where my thought about the empty query came from.

I agree that in the long run it would be better to move this feature to the regular search. But it should have the same effect as it has now. That, however, sounds quite complicated and for the moment similar search does what it's supposed to.

So, what do you suggest? Should I put this change into our local layer for the time being and we'll think about a way to do similar_search via the normal search (when there is some time to spare)?

nicolasfranck:
@petrakohorst : add a checkbox in the form "/publication" thats says something like "fuzzy search" (similar to "I'm feeling lucky" from google search). When the checkbox is set, the controller /publication can set the necessary parameters (minimum-should-match and others) in the background. At least the user sees what is going on.

phochste:
I would prefer a solution where this similar_search is removed and the functionality is part of the normal search page. Adding a redirect would make the confusion bigger.

nicolasfranck:
mm, when I think about it: better to use a fuzzy cql query in the page /publication. Looks ugly, but at least everyone knows what happens

phochste:
I close this pull request. In the code I'll add something where this other url for `similar_search` is not required.

PR #545, opened 2018-08-06
Download url

nicolasfranck:
This branch is a subbranch of "content_disposition".
So better to merge this PR https://github.com/LibreCat/LibreCat/pull/525 before thinking about this one.

What it does:

* change old download route to match better (was too restrictive)
* include alternative route with less information, excluding the file_name


nicolasfranck:
@phochste : right. The old download url ":id/:file_id/:file_name" is a redirect now to ":id /:file_id". I wanted to use an internal forward, but that does not work when you're using send_file..

PR #547, opened 2018-08-07
Add a space (was: "Allow html in file description on frontdoor")

petrakohorst:
Remove template html filter for file description on frontdoor to allow for html tags in description.

And fix incorrectly placed space in data re-use license.

### DIFF
--- a/views/publication/tab_filedetails.tt
+++ b/views/publication/tab_filedetails.tt
@@ -68,7 +68,7 @@
   <div class="row">
     <div class="col-md-3 text-muted">[% h.loc("frontdoor.tabs.file_details.description") %]</div>
     <div class="col-md-9">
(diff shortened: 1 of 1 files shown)


petrakohorst:
You're absolutely right about the risks, but unfortunately we have existing publications that use both markdown and html in the file description and in the abstract.

Making this configurable seems quite complicated. So I think it's easier to make those changes in our local layer. 

I'll change this PR to include only the missing spaces. And I'll retract PR #540 and will move both of them to our layer.

petrakohorst:
One more thing, though:

If the core does not allow markdown and html in metadata, we should also remove the hint below the abstract field in the forms. There is a line saying that users can use markdown in the abstract.

PR #558, opened 2018-08-22
Make gravatar optional via config.

petrakohorst:
This is an unintrusive change. If you don't set the config param in your local.yml, you will see no difference.

In our live system I have configured our local layer to not use Gravatar, but to use a dummy image instead. Now one of our supported institutions has also asked how to remove the Gravatar functionality. I feel like this should be configurable, rather than having to copy yet another file to the layer.

The config param has a somewhat complicated name, but this way the change is unintrusive.


petrakohorst:
I can make that change, no problem. But: then anyone NOT having set that parameter specifically will not see the Gravatar feature anymore after this PR has been merged.

Is that okay? If so, I'll change it.

(I named it in this double negative in order to not break any functionality.)

petrakohorst:
@phochste Have you seen my question regarding your requested change on this PR?

petrakohorst:
@phochste I changed it to be called "show_gravatar" and I set a default to 1 (in the template), so that Gravatar will be used even if the config parameter is not present.

PR #591, opened 2018-09-18
show correct dc:rights statements

vpeil:


vpeil:
The 'oa' field exists in the index! See `fixes/index_publication.fix`. And we removed the `file.*.open_access` field, not the oa field in the index, which is used for search/facetting.

PR #605, opened 2018-10-04
No splat

nicolasfranck:
I'm not sure about this line (from the original):

```
get qr{/project/*}
```
in lib/LibreCat/App/Search/Route/person.pm

but it doesn't seem necessary. Am I right?


nicolasfranck:
Test t/www/project.t failed during my first commit:
```
Failed test 'GET /project/PDR07/93'
Failed test 'GET /project/'
```
The latter I understand. I added a fix for that.
The first I don't: why do we need a "/" in an id, a crucial part of the url? It is strange that the trailing slash should be ignored, but a slash in the middle of an id not.

phochste:
@vpeil Can you look at this? I don't know what the required support is for project identifiers. Currently anything is allowed, which means I can't accept this pull request. Or, is there a way project identifiers can be restricted, URI encoded, etc?

PR #610, opened 2018-10-11
remove deprecated functions

vpeil:


phochste:
I put this one on hold. We have extensive layers in Ghent that still use these deprecated features. We will need a plan first how to migrate those.

PR #614, opened 2018-10-16
fix download route generated by uri_for_file

nicolasfranck:
The method uri_for_file generates these urls:

```
/download/:id/:file_id.:extension
```

but these do not work anymore. These are only used in the dropzone.

This pull request adds a route for that.


nicolasfranck:
wait before merging this

PR #619, opened 2018-10-17
move default dancer env options to config.yml; remove env

nicolasfranck:
Dancer environment files should not be added to git.
This conflicts with for example "puppet":

* puppet adds configuration file environments/deployment.yml
* puppet installs librecat, which overwrites environments/deployment.yml with default one.
* second puppet iteration (when?) fixes above problem.

What I did

I moved all options from development.yml to config.yml
Moved development.yml to development.yml.default
Moved deployment.yml to deployment.yml.default
Added environments/*.yml to .gitignore


phochste:
This is a very specific UGent use-case. I don't see any other good arguments why other institutions don't need these environment files.

nicolasfranck:
Attributes defined in environment files override those from the config.yml.
So only environment specific attributes should be put there, not the default ones.

Those defaults should be put in the config.yml

PR #623, opened 2018-10-18
can_download must do _can_do_action much sooner

nicolasfranck:
_can_do_action is only applied when status is "closed".
So files from records with status "returned" can easily be downloaded
if the access_level is open_access


nicolasfranck:
mm, seems like _can_do_action actually requires a user. So this code always returns "access denied" when accessed without login.

phochste:
@nicolasfranck This doesn't work. It blocks access to local files (e.g. UGent) for users within the IP-range of UGent.

nicolasfranck:
@phochste : yes. As I mentioned, it doesn't work as it requires a user user session.

I'll close this for now

PR #655, opened 2018-11-21
fixing 'mark all' for person's profile page

ThriceGood:
In a person's profile page, when clicking 'mark all', it is literally marking all publications in the system (to a limit of 1000). This is because there is no 'cql' param in 'searchParams'. The edit above adds the cql param if missing.


phochste:
thx!

PR #686, opened 2019-02-01
Allow users to save their citation style for the public list

petrakohorst:
Allow users (and all other roles) to set and save their prefered citation style for their own public list. Also allow delegates to set and save the style for each of their entrusted users' lists.

The setting can be made in the dropdown menu user_settings in the topbar. Delegates will see the heading "Public Citation Style (name_of_the_current_person)" above the menu to make it clear for whom they are changing and saving the style. This display and functionality depends on the selection of delegate tab they are currently on. If they change the tab to edit publications for a different person, this menu will change to the other person's style.
All other roles will see "Public Citation Style (my_own_name)" and can only change their own citation style.

Somehow the option to save a style for public display had been lost somewhere. But I'm not even sure where that was possible before...

(The route in the person.pm needed a change, too, because for some reason the Perl command "keys" does not return the whole list of keys from the hash, but only one. No idea why!)


phochste:
@petrakohorst  I have reverted this pull request. I don't think it can work. You store the long name of the citation style but the short name is needed to create the preferences form (in user settings).

As special question. Can you update the code to support POST requests to update user preferences instead of GET requests?

PR #690, opened 2019-02-08
space added to the end of characters in hexmap

ThriceGood:
example of citation with space after Š character:

`Hilbe C, S imsa S , Chatterjee K, Nowak M. 2018. Evolution of cooperation in stochastic games. Nature. 559(7713).`

after fix:

`Hilbe C, Simsa S, Chatterjee K, Nowak M. 2018. Evolution of cooperation in stochastic games. Nature. 559(7713).`

Is this my change generating the expected or was there some reason for the space?

PR #694, opened 2019-02-19
Fix extra space in RTF export after hex conversion

petrakohorst:
What this fix does:

Don't add a space after hex character conversion in all cases. A space is only needed, if a character has been converted into something like ```\\ldblquote``` or ```\\bullet```. In these cases the space can be added in the hex table directly and is not needed later in the regex.

Also don't convert special characters into some hex representation of standard characters, but convert them into those characters directly. (I.e.: don't do ``` "152" => "\\'4f\\'45"``` but do ```"152" => "OE"```)

And add two more quotation marks to the hex table to be disposed of.


vpeil:
This PR makes #690 obsolete.

PR #699, opened 2019-02-21
Fix a server error on person page (after a search)

petrakohorst:
The search on ```/person/[id]``` pages throws an internal server error. The search results (```hits.tt```) on that page call ```h.all_marked```, where parameter q is a string but an array is expected.

Solution: push q to cql first and use that parameter for further processing.

### DIFF
--- a/lib/LibreCat/App/Helper.pm
+++ b/lib/LibreCat/App/Helper.pm
@@ -194,7 +194,8 @@ sub is_marked {
 sub all_marked {
     my ($self) = @_;
     my $p = $self->extract_params();
-    push @{$p->{q}}, "status=public";
+    push @{$p->{cql}}, $p->{q} if $p->{q};
+    push @{$p->{cql}}, "status=public";
 
     my $hits       = librecat->searcher->search('publication', $p);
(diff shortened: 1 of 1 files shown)


vpeil:
Go to http://demo.librecat.org/person/3?q=test

Then type someting else in the search box, e.g. "dean". You should see the error.

phochste:
Works for me in the `dev` branch

vpeil:
I can confirm that it works on `dev` for me, too.

petrakohorst:
This problem also exists on the dev branch... BUT:

I can only reproduce it if there are two or more search results. (Both on the demo server's master branch and on my local dev branch.)

One result and there is no error:   
http://demo.librecat.org/person/3?q=space#

Two results or more and the error occurs:   
http://demo.librecat.org/person/1234?q=test#   
http://demo.librecat.org/person/3?q=test#  (this one only produces the error after I have added a second publication with the word test)

Also we need to keep in mind that the sub in question (all_marked) is called by a javascript function after the page has loaded. But I have no clue, why this error happens in this way.

Issue #598, opened 2018-09-25
Support discovery of public datasets


vpeil:
https://ai.googleblog.com/2017/01/facilitating-discovery-of-public.html

PR #700, opened 2019-02-22
Schema org json ld

vpeil:
This PR introduces schema.org in JSON-LD. This is a more flexible and extensible approach than the speical attributes within the html.

The mapping can be improved/extended in the future.

And it solves #598.

### DIFF
--- a/fixes/lookup/schema-org_types.csv
+++ b/fixes/lookup/schema-org_types.csv
@@ -1,0 +1,19 @@
+type,schema-org_type
+journal_article,ScholarlyArticle
+preprint,xx
+report,Report
+conference,conference paper
+conference_editor,conference publication
+conference_abstract,conference abstract
+working_paper,Report
+book,Book
+book_editor,Book
+book_chapter,Chapter
+newspaper_article,Newsarticle
+translation,translation
+review,Review
+encyclopedia_article,encyclopaedia entry
+patent,patent
+journal_editor,journal editor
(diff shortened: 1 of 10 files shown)


phochste:
@vpeil Can you give some hints how I can see in the web interface the effects of this pull request?

vpeil:
@phochste look at the html source on the frontdoor. At the end of the page you should see

```js
<script type="application/ld+json">
...
</script>
```

with some json inside.

nicolasfranck:
@vpeil @phochste I think it needs to be documented what this change means for ones local branch of librecat. If someones has overridden a view (e.g. views/publication/record.tt), he/she would loose this functionality. 

Thanks for this change!

nicolasfranck:
@vpeil : I think the json object needs to be a regular object, not an array of objects. The use of "export_to_string" is probably the cause for this

vpeil:
yes, are there still references left?

nicolasfranck:
ah sorry, I was looking in my own junkyard

nicolasfranck:
Look for "itemscope" and "itemtype" in the views:

* views/publication/record.tt
* views/project/record.tt
* views/header.tt
* views/home.tt

vpeil:
ok, will do.

vpeil:
> I think the json object needs to be a regular object, not an array of objects. The use of "export_to_string" is probably the cause for this.

I don't understand the problem.

Issue #579, opened 2018-09-03
Compatibility with ElasticSearch version > 5.5.x


datango:
For the next debian version librecat needs to support elasticsearch > 5.5.x

[Elastic Support Matrix](https://www.elastic.co/de/support/matrix)

nics:
I've released a new version (0.0511) of Catmandu::Store::Elasticsearch that supports recent versions of es without the need for id_key or key_prefix. I'll also add some notes to this comment on how to migrate queries and the mapping.

nics:
### mapping
- the `standard` analyzer filter does nothing and gives a warning, I removed it
- `include_in_all` is deprecated, there is no `_all` field anymore, make your own instead:
```a
aAALLL
```

nics:
work on this is being done in the feature/es65_support branch. Catmandu::Store::Elasticsearch also needs some big changes (bag_index branch)

nics:
The feature/es65_support branch combined with dev release 0.9_02 of Catmandu::Store::ElasticSearch supports Elasticsearch 6.5, all tests pass.

vpeil:
That's amazing! What's the next step? Should we test this on our own dev-machines?

PR #705, opened 2019-03-07
Elasticsearch 6.5 support

nics:
This PR adds support Elasticsearch 6.5 and addresses issue #579. This is a breaking change that requires reindexing and updating your config because:
- ES types are deprecated, so LibreCat types (and their bags) are now mapped to their own index
- the ES mapping and query dsl has changed


vpeil:
@nics can we make the host configurable? Some people might want to run ES on a different host.

nics:
isn't it already throught the store config?

vpeil:
yeah, seems to be so. I remember one use case in the previous version, where the indexing commands where host agnostic, but maybe I mixing things up. just ignore my comments

phochste:
Can this pull request be resubmitted from the dev branch and not the master branch? This will make it easier for me to review and follow up.

nics:
replaced by pull request #716 against dev branch

PR #709, opened 2019-03-08
Bugfix: related material kills oa flag

petrakohorst:
Handling related material does not make use of update_publication.fix and index_publication.fix. Hence, fields added at indexing time (not only the oa flag) are killed when a relation is added to a record.

This still needs to be tested a bit more. And maybe we should add a general test for this scenario?

### DIFF
--- a/lib/LibreCat/App/Catalogue/Controller/Material.pm
+++ b/lib/LibreCat/App/Catalogue/Controller/Material.pm
@@ -9,6 +9,7 @@ Helper methods for handling related material links.
 use Catmandu::Sane;
 use Catmandu;
 use LibreCat::App::Helper;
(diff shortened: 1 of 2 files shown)


phochste:
Ok. But at some point I think we need to revisit the implementation of this related materials. In the current implementation these records are hard linked: editing a record with `N` related links requires `2N` updates of records in the database. This is doable for records with a few related links, but what I found out in my archive..records with tens (or more) related links are just not editable anymore (because it takes so long to follow all the links and update the database).

There should be a way to optimise this. Uses shouldn't be able to touch that many records in one go. Certain in cases where there is no record locking, possibilities for race conditions etc.

PR #710, opened 2019-03-08
Pr/licenses and type software

petrakohorst:
This PR has breaking changes!

Licenses changed: 
- all licenses are stored in the field ```license``` (instead of ```data_reuse_license``` and ```cc_license```)
- license keys in the db now contain a version, e.g. ```cc_0_1_0``` or ```cc_by_4_0```
- more licenses altogether
- we now make a difference between licenses for normal publications, data publications or software (which licenses are available can be configured in lists.yml)

Existing data needs to be updated in the db for this PR to work. (will follow in comments)

----

New type software:

Copy of research data.   
With all the restrictions of RD (doi not editable, user can not publish, only submit).

Processes need to be checked:
- send email when submitting software record
- register doi etc.
- is file no longer editable for user once the record has been made public?
- ...


joschirr:
Sounds good.
Do you store and expose in metadata a machine readable license URI (if available, e.g. for CC-licenses) ?

vpeil:
@joschirr yes, we do.

phochste:
I see there is no documentation or tooling provided to help installation migrate their old databases to the new format. Could we start some kind of consideration of bash scripts (or anything) that have all the required steps to migrate in a safe way. E.g:

- Backup old data
- Stop services & convert all the data & start services

vpeil:
Same question here: add this intof to the Wiki?

phochste:
@vpeil Yes I need some detailed, fool proof instructions how to do this migration

vpeil:
@phochste please, have a look at https://github.com/LibreCat/LibreCat/wiki/Upgrade-Nora-to-Oscar#data-migration

phochste:
@petrakohorst @vpeil I'm also reverting this pull requests. All the I18N information is missing that explains in German/English the content of these licenses.

vpeil:
What exactly do you mean? The locale strings are there.

phochste:
@vpeil  For all the publications there need to be locale strings for cc_0_1_0 and all other licenses. I don't see them in the locale files. Just check the demo records in the frontend.

PR #717, opened 2019-03-27
remove legacy routes for next major version; if necessary for some in…

vpeil:
…stallations this can be handled by a webserver


phochste:
@vpeil can you provide a working example how a webserver could route the old urls?

vpeil:
@phochste sure. Just as a comment? Or would like this info somewhere else?

phochste:
@vpeil  See https://github.com/LibreCat/LibreCat/wiki/Upgrade-Nora-to-Oscar 

Search for TODO

vpeil:
done

PR #722, opened 2019-04-08
translate facet status

nicolasfranck:


phochste:
Missing translation of qae. And requires other code in the new EL6 branch. I'll merge this by hand

PR #726, opened 2019-04-15
Download headers

nicolasfranck:
Makes it possible the builtin http headers for downloads:

* for all kind of files
* by content-type (overrides previous)

Example configuration (see: config/files.yml):
```
filestore:
  download_headers:

    # ..for all content-types
    for_all:
      # the '%s' is replaced by the current content-type
      "Content-Type": "%s"
      # the '%s' is replaced by the current filename
      "Content-Disposition": "attachment; filename*=UTF-8''%s"

    # ..for specific content-types (overrides previous)
    for_content_type:
      "text/plain":
        "Content-Type": "%s; charset=utf-8"
        "Content-Disposition": "inline; filename*=UTF-8''%s"
```


phochste:
Thanks, great addition

PR #727, opened 2019-04-16
Introduce a metadata store api

vpeil:
I will add a swagger.yml for the API documentation.


vpeil:
@nics @phochste I have requested a review ;-)

You can check the documentation by running

```bash
$ bin/librecat generate openapi.yml
```

Then you can upload the file at https://editor.swagger.io/ and see how it looks. There are still some errors in this documentation, but the main part of the API docs is finished.

phochste:
@vitali can it be that this pull request adds some changes to the code that have no relation with the API changes? E.g. other indexation rules, changes the schema? Could you pull them out and create a separate pull request otu of that?

vpeil:
yes, that is possible. These were some minor improvements I came across while working on the API. I don't know if they will affect the tests. I'll be back with an update.

phochste:
@vpeil can you give also an example how in a webbrowser I can access the api and pass the authentication token?

vpeil:
1. I have pushed a change, which removes changes not related to the api (9be9272)

2. The api authentication was the most basic one for demonstration purposes. I agree with your points @nics .

vpeil:
We can rework the authentication as described in https://mojolicious.org/perldoc/Mojolicious/Guides/Cookbook#Basic-authentication1

@nics why not use `App::bmkpasswd` again? We already use it for user passwords.

vpeil:
@phochste suggested to move the filestore api to this place, too.

nics:
some small things left to do:

- [ ]  move the /filestore api to this api @vpeil 
- [ ] version the api path @nics 
- [ ] generate openapi.yml on the fly @vpeil 
- [ ] add a /openapi.yml route @vpeil 
- [ ] create a token bag @nics
- [ ] encrypt the stored token @nics

vpeil:
@phochste please have a look at it

phochste:
Can the openapi.yml configuration and template toolkit templates be moved to a separate directory and not store them in the LibreCat home directory?

I still don't seem to be able to test this API with any Swagger tools. (e.g. the Swagger plugins for Chrome). Can it be that the openapi.yml needs to have a YAML mimeType (it is now served as text/plain).

vpeil:
@phochste I tested the chrome plugin "modify headers for google chrome". Entering the authorization token one can inspect the api easily.

PR #733, opened 2019-05-06
put audit in separate package;no use more for worker

nicolasfranck:
Why this is needed: 

* now a worker is needed in order to store messages into the bag "audit", while this actually trivial.

What I changed:

* created a separate package LibreCat::Audit that does both validation and storage. 
  * implements methods from LibreCat::Validator
  * isa LibreCat::Validator::JSONSchema: the schema is located in the package itself. All attributes are required for now (id, bag, process, action and message), and no additional attributes may be added. Suggestions?
  * It uses the same Catmandu->store("main")->bag("audit") as underlying bag
  * bag methods like "add" and "delete" are proxied to that underlying bag
  * records are validated before adding them to the underlying bag

* fixed sorting of audit records: attribute "time" is supposed to be an integer, but is often sorted as a string (using "cmp").

* tests assumed that the attribute "time" could also be a string. These tests did not fail because SQLite accepts strings even if the column is marked as integer. This does not work for other systems.

* deprecated worker LibreCat::Worker::Audit. Remaining messages will be resolved by this worker, but new messages will be sent to LibreCat::Audit directly. So you can shut this one off.

P.S. who/what actually uses the route "post /librecat/audit/:bag/:id" ?

### DIFF
--- a/lib/LibreCat/App/Catalogue/Route/audit.pm
+++ b/lib/LibreCat/App/Catalogue/Route/audit.pm
@@ -12,6 +12,7 @@ use Dancer ':syntax';
 use Dancer::Serializer::Mutable qw(template_or_serialize);
 use LibreCat::App::Helper;
 use POSIX qw(strftime);
+use LibreCat::Audit;
 use URL::Encode qw(url_decode);
 
 =head2 PREFIX /librecat/audit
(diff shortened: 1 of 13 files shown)


phochste:
@nicolasfranck  

Can you check the following procedure:

```
$ bin/librecat publication get 2737400 > /tmp/brol
$ bin/librecat publication add --log "test123" /tmp/brol
added 2737400
	(in cleanup) Can't call method "DESTROY" on an undefined value at (eval 270) line 17 during global destruction.
```

Any idea why I get the error message?

nicolasfranck:
@nicolasfranck Yes I've seen that. Trying to fix it. It goes away when I do not use the state variable anymore on line 169 .., strange

PR #741, opened 2019-05-22
Add the oa sign to records in the publication list view.

petrakohorst:
Only if one of the files is oa, of course.

Sign is displayed next to year, type etc.

### DIFF
--- a/views/hits.tt
+++ b/views/hits.tt
@@ -78,7 +78,7 @@
 [% ELSE %]
     <div class="col-md-12">
 [% END %]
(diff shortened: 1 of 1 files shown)


nicolasfranck:
"oa" is the same as access_level=open_access I guess? Better to make this explicit. "oa" is probably meant as a derived field for facets only?

vpeil:
yes, `oa` is a derived field (created at index time). But it also respects OA-URLs, so it is better to use this field instead of mutliple or statements.... I think this can be merged safely.

nicolasfranck:
@vpeil it becomes a problem when someone decides that this derived field should not be created, because he/she does not want this facet..

petrakohorst:
If someone decided not to create the oa field at index time, the oa sign would simply never be displayed in the list. It would not break any code. And if they want the oa sign, they would simply have to create the oa flag during indexing.

PR #747, opened 2019-05-22
Inactive users

nicolasfranck:


nicolasfranck:
@phochste indeed there is no documentation for whatsoever the User is expected to do. In the ticket attached to this PR, I expressed my concern where "inactive" users should be checked, and I saw that apart from that Bag.pm, the attribute inactive is in fact ignored. 

So I thought, maybe put the check in the same place? I'm not sure if the authentication layer (like in Bag.pm) is the right place to put this check. So put this in the whitelist?

PR #756, opened 2019-05-31
move sleep to hook

nicolasfranck:


nicolasfranck:
@phochste : by "stated TODO's" you mean the TODO's that were removed from the controller (see git diff)?

phochste:
@nicolasfranck This sleep was required in the front-end to see the updated /added record immediately in the result list. I expect some extra coding to mimic this behaviour in the frontend. Adding a sleep hook for e.g. `publication-edit` would again result in a sleep for front-end and backend commands (which is not the intend I guess of your coding).

nicolasfranck:
@phochste the hooks are only used in the controllers. Correct me if I'm wrong.
(quick "grep" for "publication-edit", "publication-new" .. shows you that)

The drawback is of course that you have to configure the hook "sleep" (in hooks.yml) to mimic the old
behaviour. As these hooks are empty by default, this change is not backward compatible. I saw no way to change that.

The backend commands were slow because the sleep was hardcoded into the LibreCat::Model (base class!). By putting this in a hook, only the controllers will have this delay.

PR #768, opened 2019-06-06
Pr/addendum to pr 700

vpeil:
follow up to the discussion in PR #700


nicolasfranck:
mm, the schema.org references you removed. Shouldn't they me mentioned elsewhere (json ld?), or are these not actually important?

vpeil:
I think, it is expressed via `"@context":"http://schema.org/"` within the json-ld

PR #771, opened 2019-06-06
Pr/strict web validation

phochste:
- Making year field in publication a strict `\d{4}` year. 
- Percolating validation errors to the web interface via flash messages
- This as first step to remove the JavaScript based validation.

For now the validation just adds extra syntax validations not covered by the JavaScript validators.


nicolasfranck:
@phochste Could you make the errors in lib/LibreCat/App/Catalogue/Route/publication.pm translatable, using locale?

PR #773, opened 2019-06-07
cleanup elasticsearch field mapping

nicolasfranck:
* create separate sort fields for text fields
   * subfields "exact" (e.g. title.exact) could not be used.
   * analyzed text cannot be sorted except when fielddata is set, but that has a huge memory impact. Even if the analyzer "tag" is interpreted as analyzed.
   * a possible solution is the use of type:keyword in combination with a normalizer ( https://www.elastic.co/guide/en/elasticsearch/reference/current/normalizer.html ). I created two normalizers:
     * sort_text: lowercase and ascii folding
     * sort_keyword: lowercase only
   * publication.status and publication.type still have fielddata. Shouldn't they be of type "keyword"?

* analyzer "datestring" does nothing, but for ES it is an analyzed field, so no sorting is possible. I replaced the types here by "keyword".

* create explicit field mapping:

  If you don't supply a field mapping for text fields, ES creates this mapping:

```
type: text
fields:
  type: keyword
  ignore_above: 256
```   
  So it creates two fields for each unmapped field: $field and $field.keyword

  It is also unsafe, as ES derives the type by examining the first appearance. 
  If the first document has an invalid value (e.g. integer instead of string),
  the first document is accepted, and the other (with a correct value) are discarded.

  I added the following for model publication

    * cc_license
    * project.name
    * file.file_id
    * file.file_name
    * file.relation
    * file.title
    * file.description
    * file.date_created ( was "date" instead of "datetime_no_millis" )
    * file.file_size
    * file.request_a_copy (was "long" instead of "byte")
    * file.access_level ( was "text" instead of "keyword" )

  I added these fields for model user:

    * email ( type "tag" ? Shouldn't that be "keyword"? )
    * style ( was "text" instead of "keyword" )

Some questions:

* Are all subfields "exact" still necessary? A lot of them were only used to sort, and I added
a non-stored subfield for sorting.
* what is the use of User.email_encoded?
* why do we use a analyzer "tag" for User.login? To search case insensitive?


vpeil:
@nicolasfranck the field `cc_license` is now called `license`

vpeil:
@phochste this is one is ready to merge

Issue #791, opened 2019-08-22
Plack::App::Cascade causes confusion when one app gives status 404


nicolasfranck:
Now three apps are loaded in order by `Plack::App::Cascade`:

* `Plack::App::File` (for serving static files from directory `public`)
* `Dancer` (librecat search and admin)
* `Mojolicious` (the file api)

When an app returns status code `404`, `Plack::App::Cascade` ignores
all content, and gives control to the next app.

cf. https://metacpan.org/pod/Plack::App::Cascade

For example: when the file controller from LibreCat (`LibreCat::App::Catalogue::Route::file`) 
does not find a file,
it returns status 404, and template '404'. But this content is ignored,
and all control is given to Mojolicious, that does not find the file either,
and returns the content from `views/404.html.tt2`.

See: https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App/Catalogue/Route/file.pm#L174 . The content `The file does not exist anymore. We're sorry.` is never shown.

I think two fallback levels is the problem: only the first two apps should exist.
The Mojolicious should be excluded from `bin/app.pl`. It can be started
separately anyway.

PR #792, opened 2019-08-23
Endpage content

nicolasfranck:
See https://github.com/LibreCat/LibreCat/issues/791


vpeil:
ok, but then we need some startup script for the Mojo app, e.g. `bin/mojo-app.pl`

Plus some documentation how to set this up with an apache proxy.

nicolasfranck:
@vpeil there is no need to create an extra startup script (as boot.sh), as this is the way to start a plack application:

```
plackup bin/mojo-app.psgi -p 5000
```

vpeil:
ah, great. Haven't noticed the one, sorry.

vpeil:
@nicolasfranck I think, we need to revert 95a0d42 to make that PR consistent.

nics:
the mojo app should just be removed from the cascade mounted under '/api' in bin/app.pl, that also solves the problems, no need to have a separate startup script for mojo.

```
    mount '/api' => $mojo_app;
    mount '/' => $dancer_app;

```
these lines should also be removed from LibreCat::Application:
```
    push @{$self->renderer->paths}, @{librecat->template_paths};

    $r->any(
        '/*whatever' => {whatever => ''} => sub {
            my $c        = shift;
            my $whatever = $c->param('whatever');
            $c->render(template => '404', handler => 'tt2', status => 404);
        }
    );
```
sharing templates between mojo and dancer is problematic (as witnessed by the warnings when running the tests)

nicolasfranck:
@vpeil @nics ok will fix this in branch fix_api_filestore, as it is related

vpeil:
@nicolasfranck @nics I have pushed changes as suggested above.

nicolasfranck:
@vpeil wait,  I've prepared a working bin/app.pl in another which I'll merge once I fixed some more internal work. I'll end this pull request for now. Sorry.

PR #801, opened 2019-09-11
fix typo

nicolasfranck:


phochste:
I will do this manually. This is not only a typo but this method should mix $p and p at all...

Issue #797, opened 2019-09-09
Inconsistencies in schema/code


vpeil:
* date_deleted is set in code, but not defined in schema:
https://github.com/LibreCat/LibreCat/search?q=date_deleted&unscoped_q=date_deleted

what is the use of this field? Eithter remove it or extend the schema?

vpeil:
Do we want to store `date_deleted`? Any opinions on that?

phochste:
@vpeil I'll extend the schema

PR #804, opened 2019-09-13
move fields from store to index due to stricter checks of the validator

vpeil:
Since version 2 the validation is much stricter (which, of course, is better). Unfortunately some inconsistencies of the past appear. This tries to clean things up:

- Move derived flags to `index_publication.fix`: oai_deleted, isi, pmid. These are set during indexing time now.

- project.call_identifier and project.grant_number are not allowed in schema, so move this to index, too

- deparmtent.tree not in schema, so add this at indexing time

See also the related question in #797

### DIFF
--- a/config/schemas.yml
+++ b/config/schemas.yml
@@ -605,18 +605,6 @@ publication:
             oneOf:
               - type: string
               - type: integer
-    pmid:
-      oneOf:
-        - type: string
-          pattern: "^[01]$"
-        - type: integer
-      description: "Flag of PubMed"
-    isi:
-      oneOf:
-        - type: string
-          pattern: "^[01]$"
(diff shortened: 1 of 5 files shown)


phochste:
@vpeil Ok, but I don't want to change the schema.yml which will make all records in current version 2 setups immediately invalid. Could we keep the pmid/isi field in the schema but flag them in documentation as deprecated (generated on the fly?)

phochste:
@nicolasfranck @nics do you need the `department.tree` for your use cases somewhere in the `main` database or is to have these fields in  the `search` database good enough?

nicolasfranck:
@phochste nope, go ahead and nuke that stuff

PR #806, opened 2019-09-16
Fix api filestore

nicolasfranck:


vpeil:
Documentation is missing for that part of the API (again as openapi spec)

nicolasfranck:
Mm, wait before merging this. Are we sure we want the filestore as a separate route?

`/v1/file` suggests that files are managed separated from its record, which is (and should not) be true. That's why editing of records is disabled here for now.

Better?

`/v1/publication/:id/file`

Then in the future one can move the whole file handling to model LibreCat::Model::Publication ..

PR #807, opened 2019-09-18
fix bug in cloning functionality

vpeil:
Due the the '_version' field, which shouldn't be present in case of a new
record. Same for the 'date_created' and 'date_updated'; besides the date plugin
in the store will take care of that.


vpeil:
will push changes in a minute!

vpeil:
Please, have a look at #819 .

I'll close this one.

PR #813, opened 2019-10-03
fix decode_file in hook publication_file_sort

nicolasfranck:
Before this hook generated an error when you upload a single file,
because it did not expect a stringified object.


nicolasfranck:
When using my own hook `LibreCat::Hook::publication_file_sort`,
I stumbled on a problem: in a "before" hook `record.file` can look
like this:

* a single json string, when one file was uploaded
* an array of json strings, when multiple files were uploaded
* undefined

This behaviour is due to the file uploader, and the way dancer reads request parameters.
So this needs to be handled before any hooks are applied (the method nested_params does this also!).

What I did:

* moved the method `_decode_file` to a separate hook (should not be used otherwise)
* remove method `_decode_file` from `LibreCat::App::Catalogue::Controller::File`
* apply the hook in route `/publication/update`, and added a reminder to remove this in the future.

Another thing I saw, but could not fix: when you save a record without adding files, the attribuut `file` disappears.

phochste:
@nicolasfranck  I understand what you are trying to create. But, can you tell me why did you put the `_decode_file` into the Librecat::Hook namespace? Hooks are meant to be code that are run optional (like features you want or not). Isn't this something more of a Helper function

nicolasfranck:
@phochste you could also add the method _decode_file directly into the route. At least this decoding should be done before any hooks are applied. But you're right about the way I treated this.

PR #821, opened 2019-11-05
Pr/mojo search api

vpeil:
This PR introduces new search api routes `/api/v1/seach/$model`.

They are protected by the same token as they are under `/api/v1`.

The JSON-Response contains the hits, the aggregations and the pagination information, so it can be used in the future by a separated frontend.


vpeil:
> Hi, some remarks:
> 
> * I would expect /search to be an operation on the model, why not just add it to the model api?
> * I would make a Searchable plugin with a search method for the model (cfr. Versioning) that for now just dispatches to the Searcher. It's a much better encapsulation and code organization if all data operations go through the model.
> * the model api could autodiscover if the model is capable of searching (cfr. Versioning)

Ok, I'll try update the code based on your suggestions.
> 
> And some questions:
> 
> * do we want to expose the aggrgations in their raw form, that ties the api to this version of elasticsearch?

hmm, that ist probably the better solution.
> * do we want to accept anything else than cql as query?

Right, cql only is better.

phochste:
@vpeil @nics can you confirm that versioning is already part of this API (and other API we already have)

nics:
The route contains the version, is that what you mean?

phochste:
@nics yes that is what I mean: there is a mechanism to deprecate the current APIs somewhere in the future?

nics:
yes, that can be used as a very simple way to do that

vpeil:
@nics @phochste 
Commit 53b494b7dc832c5faaef118c90a53ae9672d7fda introduces a parameter to switch on aggregations. Default behaviour is not to export aggregations.

After checking, I think this PR is ready for another review or a merge.

phochste:
Documentation need, something is working but I don't know how to test it.

vpeil:
in light of PR #881 is this PR obsolete. There search API should be handled in a follow-up to PR #881.

PR #822, opened 2019-11-05
cleanup mojo application and mount differently

nicolasfranck:
* mount dancer and mojo under different path with different rules:
  * dancer: session based
  * mojo: open or token based
* remove references to api_access from config and LibreCat::App. Note that certains paths in the api must be protected by the frontend server now (which is more efficïent), while this was done by librecat. 
* default route in mojo should return json, not using a template system.

This should also fix https://github.com/LibreCat/LibreCat/issues/791


nicolasfranck:
note done yet, sorry

phochste:
@nics I can't accept a pull request that will not have an IP check on the API. Can there be a solution where the API is started on a different port?

nics:
comment is meant for @nicolasfranck?
why can't you have an ip check on a path?

nicolasfranck:
Ok, I will remove mojo app from bin/app.pl

vpeil:
I don't understand the problem. Isn't it enough to protect the api with the authorization header?

vpeil:
Mount mojo application in bin/app.pl, please.

PR #826, opened 2019-11-06
Make worker command more robust

nics:
- switch from Gearman::XS to Gearman client
- rewrite of daemonizing code

The queue Perl api and config doesn't change, but the command line interface changes a bit:

`librecat worker start`: starts all the workers in the config. There is no stop command, info is in the worker command synopsis.

`librecat queue`: removed the start and stop subcommands


vpeil (commented):
The command `librecat worker start` hangs.... after CRTL-C the workers seem to be started as `bin/librecat queue status` returns the expected output. Any idea why it's hanging?

nics:
`librecat worker start` starts it as a foreground process, so that's expected. But ctrl-c should also stop the associated worker processes, can you check with `ps -elf -o comm | grep libre`.

vpeil:
you are right! Status printed the worker, but as not running! My fault.

vpeil:
We need to modify the scripts in `etc/`, too.

vpeil:
@nics what to do about this one since we are moving from Mojo away (for the api)? Can this be still used a standalone solution (minion only)?

vpeil:
What are we doing with this PR? It's a big change + conflicting files appeared in the meantime.

nics:
Hi @vpeil, sorry don't have time to work on this anymore. I would just close it and keep the branch in case someone else wants to work on it.

vpeil:
Yes, that's totally fine in view of our re-calibrated strategy. I would vote that same way.

PR #827, opened 2019-11-06
Fixes #794 : star mandatory added to DDC field for each new upload

### LINKED ISSUE #794: "starMandatory" added to label of field "DDC" for every uploaded file
When you upload 10 files,
10 "*" are appended to the label of field.
If the upload is slow, you actually see it appending..

Screenshot 2019-09-06 at 11 38 23

Funny bug by the way

petrakohorst:
Solution: check via jquery if star is already present and only add if it is not.


phochste:
@petrakohorst thx!

PR #840, opened 2019-11-21
Change file deletion popup message

petrakohorst:
When the user wants to delete an uploaded file in the edit form, a message is displayed, saying that the user can edit/re-upload a different file. That is not possible anymore. I changed the text accordingly.

### DIFF
--- a/public/javascripts/librecat.js
+++ b/public/javascripts/librecat.js
@@ -520,7 +520,7 @@ function edit_file(fileId){
  * @param fileId = file ID
  */
 function delete_file(fileId){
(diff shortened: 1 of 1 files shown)


phochste (changes requested):
```
Are you sure you want to delete this uploaded document? 
Any external links will be broken!
(If you want to upload a new version of your file, please create a new record.)

Delete this file?
```

Wouldn't it be better to make this message into local I18N version. I don't think one agree that
creating a new record is a good advice in all possible LibreCat environments.

petrakohorst:
@phochste I agree with you on both counts: that not all LibreCat environments would like to handle file uploads this way and that this text would be better placed in a locale file.

Unfortunately this a js file where no l18N processing or templating takes place.

It is no problem for us to keep this change within our own system, though. I'll close this PR.

PR #843, opened 2019-11-29
cleanup the API namespace; FileStore not use and will be included int…

vpeil:
…o mojo api; livecitation is not an api, but some sort of mixed content and not necessary at all

Considered as a follow-up to PR #821


vpeil:
The main motivation of this PR is to use the "API" notion consistently within the code.
So, only the mojo api code will be considered a real API (and fully working). The rest is not an API of LibreCat (like oai or sru), but integrates other those API from other code bases or the is an awful mixing (like the citation "api").

nicolasfranck:
@phochste this pull request removes the unused api's, but it also removes the automatic loading of all routes under the namespace `LibreCat::App::Api::*`. Your repo "grep" depends on that behaviour. Better to add a mock package, and keep the automatic loading (but deprecate it)?

vpeil:
> @phochste this pull request removes the unused api's, but it also removes the automatic loading of all routes under the namespace `LibreCat::App::Api::*`. Your repo "grep" depends on that behaviour. Better to add a mock package, and keep the automatic loading (but deprecate it)?

@phochste what path should we follow? As suggested: Mock package + deprecation?

vpeil:
@phochste thinks this is ok.

PR #844, opened 2019-12-03
Feature/deduplication

vpeil:
This PR introduces a new class `LibreCat::Dedup` and a simple implementation of this class for publication data.

How it works:

- In the Web-UI a dedup check is performed based on identifiers only for the moment.
- In case a duplicate is found the user gets a link to the existing record.
- In the opposite case the import is peformed as before.


vpeil:
- turn idea into hooks
- don't change current behaviour

vpeil:
hook `check_publication_duplicates` added. Writes information into message field.
This also gives a blueprint if one wants to change the behaviour: copy the hook to your local layer and do what you want.

Issue #718, opened 2019-04-01
support of OpenAIRE Guidelines v4


joschirr:
Hi,
is there any chance to have support for OpenAIRE Guidelnes v4, released in November-2018

* https://openaire-guidelines-for-literature-repository-managers.readthedocs.io/en/v4.0.0/introduction.html
* http://dx.doi.org/10.5281/zenodo.1299203

which introduces a new application profile build on elements from DublinCore, DataCite and OpenAIRE, makes use of COAR Vocabularies and integrates with identifier systems such as ORCID and FundRef.

vpeil:
Hi @joschirr, if it's not a april's fool, I will definitely have a look at it.

vpeil:
@joschirr question w.r.t embargo data:
The [documentation](https://openaire-guidelines-for-literature-repository-managers.readthedocs.io/en/v4.0.0/field_embargoenddate.html) does not allow repeated fields here (except for the case "accepted" and "available"). In our data model one can set the embargo date for every file. Or should we consider this output if and only if **one** file exists?

joschirr:
embargo is considered for the whole record rather than on the file level.

vpeil:
But that's not an answer to my question. Just take the first file with embargo then?

joschirr:
so to say. Or the file with the earliest embargo start date and the longest embargo end date

PR #845, opened 2019-12-05
Pr/issue 718

vpeil:
Solves #718.

+ extends OAI config with all publication typs as sets (and better grouping)


phochste:
@vpeil I suspect the software part of this pull request for OpenAIRE v4 support will be working. But we feel some concerns here in Ghent on the interpretation of the mapping. E.g. to suddenly map 'review' records as 'book review' and 'conference_abstract' as 'conference poster' and putting all these things in a COAR tree is open for a whole lot of interpretation of the data. Do you see that there could be a bit of "uncomfort" to reinterpret the data this way? 

Here in Ghent we use our own OAI port to provide data, so we have are own choice if we want to follow OpenAIRE v4 or not. Just to check with the group in want way we can best add this to the code base. Is this mapping doable as first start?

vpeil:
@phochste we should discuss this with @joschirr this week.

vpeil:
use more broader type mapping instead of lying ;-)

vpeil:
I have pushed an update with a more general type mapping.

PR #849, opened 2019-12-24
support schema draft 07

nicolasfranck:
json schema draft 07 supports if-then-else (sub schema).
This way we can add some conditional validation into the
model. I'm not sure, but is there some conditional validation
in the frontend now, and that we can/should get rid of?

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -36,7 +36,7 @@ requires 'Catmandu::OAI' , '0.16';
 requires 'Catmandu::RIS', '>=0.04';
 requires 'Catmandu::SRU','0.039';
 requires 'Catmandu::Store::ElasticSearch', '>=1.0';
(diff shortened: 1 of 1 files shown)


nicolasfranck:
Sigh, the tests are failing because of an unrelated test: t/Catmandu/Plugin/DynamicChecksum.t
Anyone an idea what happened here?

nicolasfranck:
So wait for https://github.com/LibreCat/LibreCat/pull/850

nicolasfranck:
but I'm a bit confused about why in continuous-integration/travis-ci/pr that plugin test does succeed.

vpeil:
@petrakohorst @nicolasfranck I think there are conditional validation checks in the frontend handled by a javascript.

PR #854, opened 2020-01-07
Non-breaking change: add an option to internal_view route

petrakohorst:
Make it possible to select the search index as data source for internal_view instead of db.

This PR does not change the default behavior!

If the parameter "searcher" is added to the internal_view URL, the data will not be coming from the db (pure data) but from the search index (enriched with index fields and fixed).


nicolasfranck:
Better not
`
publication->search_bag->get($id)
`
instead of
`
$hits = searcher->search('publication', {cql => [ "id=$id"], limit => 1});
$rec = $hits->{hits}->[0] if $hits;
`
?

No need for any  cql here.

petrakohorst:
@nicolasfranck Yes, indeed! That's much better. Thanks for the hint.

Will push changes to this PR.

PR #855, opened 2020-01-07
separate use route params from query params

nicolasfranck:
* `params` in Template Toolkit is not a method like in Dancer, because Dancer adds the result from `params()`. So I added these tokens (not as a method):

  * `params_query`
  * `params_body`
  * `params_route`

* I replace `request.params.hash` in the views by `params_query` (which is always a hash)

* question: still necessary to directly use `config.locale.$lang`? Better not use `h.locale()` everywhere? I left it alone, but it is widely used in all field generators. Makes it hard to be replaced in this case.

* `div` with id `language_id` still necessary?

* `qp` is sometimes the query parameters, and sometimes `h.extract_params`. What is the difference?

### DIFF
--- a/lib/LibreCat/App.pm
+++ b/lib/LibreCat/App.pm
@@ -74,6 +74,17 @@ hook before => sub {
     }
 };
 
+hook before_template_render => sub {
+
+    my $tokens = $_[0];
+
+    #params in TT is hash resulted from call to "params()"
+    $tokens->{params_query} = params("query");
+    $tokens->{params_body} = params("body");
+    $tokens->{params_route} = params("route");
+
+};
+
 sub _login_route {
     my $conf = shift;
     sub {
--- a/views/admin/account.tt
+++ b/views/admin/account.tt
@@ -1,7 +1,6 @@
-[% qp = request.params.hash %]
-[% qp.delete('splat') %]
+[%- qp = params_query -%]
 [% lang = h.locale() -%]
-[% backend = request.path_info.match('librecat') ? 1 : 0 %]
+[%- backend = 1 -%]
 
 [% INCLUDE header.tt %]
 
@@ -21,7 +20,7 @@
 	<div class="col-md-5">
 	  <label for="id_q"><span class="fa fa-search"></span> Search by name or ID</label>
 	  <div class="input-group">
-	    <input type="text" class="form-control" name="q" id="id_q" placeholder="Einstein, A" value="[% qp.q %]">
+	    <input type="text" class="form-control" name="q" id="id_q" placeholder="Einstein, A" value="[% qp.q | html %]">
 	    <span class="input-group-btn">
 	      <button type="submit" class="btn btn-default"><i><span class="fa fa-search"></span> </i>Search</button>
 	    </span>
@@ -31,7 +30,7 @@
     </form>
 
     [%- IF error %]
-      <p class="text-danger">[% error %] <a href="[% uri_base %]/librecat/admin/account/edit/[% qp.id %]"><span class="fa fa-pencil fw"></span>Edit</a></p>
+      <p class="text-danger">[% error %] <a href="[% uri_base %]/librecat/admin/account/edit/[% qp.id | uri %]"><span class="fa fa-pencil fw"></span>Edit</a></p>
     [%- END %]
 
     <div class="row col-md-12 margin-top1 margin-bottom2">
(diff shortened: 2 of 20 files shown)


vpeil:
> - div with id language_id still necessary?

I don't think it's needed. I can't find any reference to these id elsewhere.

vpeil:
> - question: still necessary to directly use config.locale.$lang? Better not use h.locale() everywhere? I left it alone, but it is widely used in all field generators. Makes it hard to be replaced in this case.

A good point. It most cases that can be replaced. There were 2 or 3 edge cases with some problems, but can't remember exactly. This should be handled by a separate PR.

nicolasfranck:
> > * question: still necessary to directly use config.locale.$lang? Better not use h.locale() everywhere? I left it alone, but it is widely used in all field generators. Makes it hard to be replaced in this case.
> 
> A good point. It most cases that can be replaced. There were 2 or 3 edge cases with some problems, but can't remember exactly. This should be handled by a separate PR.

I will, and I'll leave the (implicit) variable `lf` (example: https://github.com/LibreCat/LibreCat/blob/master/views/backend/generator/master.tt#L4) in place so old field generators in someones layer still work, but I'll replace the `lf` by `h.loc` in the existing code because it delivers the same results

nicolasfranck:
> > * div with id language_id still necessary?
> 
> I don't think it's needed. I can't find any reference to these id elsewhere.

done: https://github.com/LibreCat/LibreCat/pull/855/commits/8868a15383437832747f2fa4446c50feb102f5ad

petrakohorst:
Loading the language specific words from the config files into a variable is necessary in one place (as far as I know... there might be more): when embedding iframe or js.

We've had problems with embeddings that specify a lang parameter (some people embed a list in english on one page and the same list german on another one). In that case I have had trouble loading the specific language via h.loc and have instead loaded the requested language file into a variable. This also concerns the links.tt file for example, as that is embedded along with the hits.

But that does not concern forms or anything else in the backend!   
Anything backend-specific can (as far as I can see) be changed to h.loc, which makes life a whole lot easier!

nicolasfranck:
> Loading the language specific words from the config files into a variable is necessary in one place (as far as I know... there might be more): when embedding iframe or js.
> 
> We've had problems with embeddings that specify a lang parameter (some people embed a list in english on one page and the same list german on another one). In that case I have had trouble loading the specific language via h.loc and have instead loaded the requested language file into a variable. This also concerns the links.tt file for example, as that is embedded along with the hits.
> 
> But that does not concern forms or anything else in the backend!
> Anything backend-specific can (as far as I can see) be changed to h.loc, which makes life a whole lot easier!

`h.loc("help")` gives the same result as `h.config.locale.en.help` if the current locale is `en`.  `h.locale` returns the default language code (seems to be `en`, always). In case of your embed use case, `lang` is set to the default locale `en`, unless the user has visited the librecat website before and has changed the language.

A part of the problem is of course the fact that there is no reference to a language in the url,
so you can only set the language by visiting the original website, and change the language there.

nicolasfranck:
@petrakohorst regarding the combination of localization and embedding, the reason why it did not (and still does not) work is the following:

* the current locale is reset, not only by `/set_language`, but also implicitly by `Dancer::Plugin::Lexicon`, that uses the key `lang` either from the query parameters or the session. So every parameter `lang` changes the locale permanently (until it is changed again).

* If you would include `lang` as query parameter in the embed link, then yes it would work, but you would also unwillingly setting the language of the librecat website. Imagine visiting an English website that uses the embed code (with of course the language set to English), and afterwards you return to the original librecat website, only to see that the language had suddenly changed from German to English.

What should happen:

* Either the locale should only be set by the query parameter, and not remain in the session. The config option `show_locale` does not change that: the session is still infected due to the way the plugin works (`/set_language` is actually not necessary).

* Either use a subpath that is not shared by the same session, and so keep the internal and external behaviour separate.

nicolasfranck:
@petrakohorst ok it seems that the problem with this stubborn language behaviour is actually my fault ;-). See https://github.com/LibreCat/LibreCat/blame/master/lib/LibreCat/App/Helper.pm, and look for the `hook before`.

PR https://github.com/LibreCat/LibreCat/pull/863 should fix that

phochste:
@nicolasfranck running Cypres integration tests I see missing #language_id errors in the `/record` , `/department` and the homepage. What does this mean and is it problematic?

PR #859, opened 2020-01-16
Fix export

nicolasfranck:
What I've changed:

* format `jsonl` should be line delimited. I've set content-type to `text/plain` as it seems more appropriate. Beware: some json browser plugins still try to parse the json, so this will only succeed if you have exactly one record...

* format `mods`:
  * removed option `noRoot` (not used)
  * removed option `xmlHeader` (exporter option `xml` does that already)
  * added template_before and template_after that wrap the results in a `modsCollection`.

* format `rdf`
  * removed option `headless`: everything is a collection so template_before and template_after should take care of that.

Questions/TODO:

* format `ris`: command `curl` ends with `5 bytes remaining to be read`. Not sure what happens here.
* format `datacite`: does not seems to have a collection wrapper. Everything has to be a list, so can we still support this? Wrap it in a OAI-PM element?
* function `LibreCat::App::Search::Route::export::_export` wraps the exporter with an `eval`, ands responds to an error with http status 404. What is intented here? What about server errors, that should be status 500?
* There is a bug in `Catmandu::Exporter::Template`: when you have 0 hits, then only the content from `template_after` is in the response. Reason: `template_before` (and xml header also) is only added during a call to `add`


vpeil:
> format ris: command curl ends with 5 bytes remaining to be read. Not sure what happens here.

Cannot confirm: `wget` and `curl -O` work fine. I think you need he option is important here, because we are expecting a file, right?

nicolasfranck:
TODO: https://github.com/LibreCat/Catmandu-Template/pull/11

nicolasfranck:
@vpeil 

```
curl "http://demo.librecat.org/export?limit=10&fmt=ris"
```

returns

```
curl: (18) transfer closed with 2 bytes remaining to read
```

on stderr

nicolasfranck:
@vpeil what to do with format `datacite`?

vpeil:
> function LibreCat::App::Search::Route::export::_export wraps the exporter with an eval, ands responds to an error with http status 404. What is intented here? What about server errors, that should be status 500?

The eval block has been introduced by @phochste (https://github.com/LibreCat/LibreCat/blame/master/lib/LibreCat/App/Search/Route/export.pm)

vpeil:
> @vpeil
> 
> ```
> curl "http://demo.librecat.org/export?limit=10&fmt=ris"
> ```
> 
> returns
> 
> ```
> curl: (18) transfer closed with 2 bytes remaining to read
> ```
> 
> on stderr

I looks like an error on the demo server (apache setting?). I can't reproduce that neither on my local dev box nor on my production system.

phochste:
I can confirm this bug locally on my dev instance when adding some records with some Russian,Chinese text...

nicolasfranck:
@vpeil @phochste about the ris export: dancer was to be blamed. Dancer tries to reencode data based on the given content type. Now it overrides everything to prevent this behaviour, and it encodes the data itself.

nicolasfranck:
format datacite is now wrapped in OAI-PMH, and content type set to "application/xml".
Not sure about that last one.

nicolasfranck:
can be merged

vpeil:
remove datacite from the route /export, but keep template for the worker

nicolasfranck:
@vpeil : done

PR #863, opened 2020-01-24
Fix auto locale

nicolasfranck:
Two things:

* I added a dancer hook in the past that changed the language in the cookie lang based on the query parameter "lang". But there is already a route /set_language for that, and it makes it hard to override the language temporarily.

* Set default cookie setting to Strict

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -18,6 +18,7 @@ session_state:
     ## Turn in in https installations
     # secure: 1
     expires: 1209600
+    samesite: Strict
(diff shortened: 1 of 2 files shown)


nicolasfranck:
@phochste fix on the way. Apparently outside of a web context, a call to `request()` returns 
`undef`. That's why `params()` fails (which is a shortcut for `request()->params()`.

But there are other calls to `params` in that file too ..

PR #866, opened 2020-01-28
Fix handling of authors and editors in crossref mapping fix.

petrakohorst:
Crossref keeps adding fields to authors/editors, which then upset the validator and kill the import. Instead of removing fields one by one, just add needed fields to tmp and discard everything else.


petrakohorst:
I have added another commit to this PR.

Adding some additional fields for crossref imports: volume, issue, publisher, alternative title, keyword, article number.   
And add a vacuum() at the end to get rid of empty fields coming from crossref.

PR #869, opened 2020-01-31
change search functionality in website to "dismax" like queries

nicolasfranck:
The current implementation of `LibreCat::Search->search` does the following with the query parameter`q` in routes like `/record`:

* split on whitespace
* construct a cql query on each part, using the default index
* append the cql queries from the parameter `cql` to this list
* join all these queries in one cql query with the operator `AND`

So you get a `cql_query`. 

I've decided to let ES do the parsing for parameter `q`, and only use cql for the filters.
This way we have more fine grained control over the search functionality and the boosting,
and there is no need to push search requirements into the cql parser.

I'm using the query type `simple_query_string` by default, which almost corresponds to the Solr dismax query, but you can turn off some parser options. All is hardcoded for now, but should be configured (e.g. using a different list of fields instead of just "all").

The `cql` parameter is now translated into an ES filter, which means that it has no influence on the score anymore (filters never should).

This also has implications for routes `/librecat/admin/account`, `/librecat/admin/project`, and `/librecat/admin/research_group` as they are using that same method.

The changes to project.pm and feed.pm were a consequence of this change: the parameter `q` should be a string, and not an array of strings (why was that?).


nicolasfranck:
btw, does anyone know where this part is used for: https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/Search.pm#L42

vpeil:
> btw, does anyone know where this part is used for: https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/Search.pm#L42

I think this has to do with templating. The bad part is that we pass `$rec` to templates and not `{rec => $rec}` (this change wo require some work).

nicolasfranck:
Similarily one should tackle the search functionality in `LibreCat::App::Search::Route::ajax` where any term can easily break/crash the application

nicolasfranck:
So again, is this acceptable?

vpeil (approved):
I think, I've already approved it yesterday.

Issue #830, opened 2019-11-07
Separate citation feature from LibreCat


vpeil:
This would benefit our overall strategy to split this application into smaller chunks.

1. It is a optional feature. A lot of use cases do not need it. 
2. It can be done in a separate repository.
3. Citations should be stored in a separate database. Switching to new styles, improving the CSL mapping, etc. would then **not** result in updated records in the main store.

vpeil:
See PR #877

phochste:
PR #877 was closed with unmerged commits.

PR #877, opened 2020-03-02
Remove citation feature

vpeil:
# What is does:

- Removes the citation feature from this repository as proposed by #830 
- Functionality is contained in a separate layer: https://github.com/LibreCat/librecat-citation
- Activating this layer should lead to the same result as before. 

# How it works:

- With the librecat-citation layer active a hook triggers the generation of citation string at `publication-update`. These citations are stored in a separate bag and get merged into the index record only.


vpeil:
I removed the custom RTF exporter because I found a better solution via pandoc: this is contained in https://github.com/LibreCat/librecat-citation

One can export publication lists in `docx` and `odt` formats now.

phochste:
@vpeil what is the status on this aparently there are still references to the citation engine in the code. E.g. `publicaton/record.tt` as was found out by @nicolasfranck . I don't know if there are more bits and pieces?

nicolasfranck:
@phochste I and Vitali have agreed to split those templates into smaller pieces, and to include those pieces, controlled by configuration. Of course, it is kind of ridiculous to add a configuration option to include a template in `views/topbar.tt`.

vpeil:
There are a lot more places where citations are mentioned. So it would be easier for the moment to leave it as is.

PR #881, opened 2020-03-16
Mojo remove

nicolasfranck:
We've (@nicolasfranck  and @nics ) decided to skip Mojolicious for now, and put all controllers back into Dancer. In the future it will be better to migrate to Mojolicious all together instead of creating a situation in between.

Mojolicious contained three controllers:

* Api.pm (i.e. "openapi")
* ModelApi.pm
* FileApi.pm

The last one was not finished, so I removed that one for now.

All controllers are now back in the namespace `LibreCat::App::Api::`:

* LibreCat::App::Api::openapi
* LibreCat::App::Api::json_api_v1

Yes, I've decided to rename the model_api to a more specific name.

Regarding `openapi.pm`:

* a "before hook" disables storage of generated session cookies. Cookies are still sent though.
* all else is left unchanged.

Regarding `json_api_v1`:

* a "before hook" disables storage of generated session cookies. Cookies are still sent though.

* a "before hook" adds IP based authentication

* a "before hook" adds JWT based authentication. I noticed that in the previous version the token was simply used as value in the header `Authorization: my.jwt.token`, which is wrong. One should use `Authorization: Bearer my.jwt.token`.
* a "before hook" requires correct content negotiation, i.e. clients should add the header `Accept: application/vnd.api+json`, and also `Content-Type: application/vnd.api+json` when posting data.

* tried to use the same names as much as possible:
  * config option for JWT token is moved from `api.v1.secret_token` to `json_api_v1.secret_token`
  * config option for ip ranges now in `json_api_v1.ip_range`
  * hooks now use the prefix `json-api-v1-`. e.g. `json-api-v1-publication-create`
  * validator for request body uses localization from `validator.json_api_v1.errors` (see below)

* added routes with named parameters (e.g. `get "/:model/:id"`) to catch every possible model, instead of using the model name in the route itself (e.g. `get "/$model/:id"`), in order to reduce the number of routes.

* Routes that update things should have JSON request body that conforms to the JSON API:
  * attribute `data` is required and an object
  * attribute `data.type` is required
  * identifier should be put in `data.id`.
  * record should be put in `data.attributes`.

  => in the previous version, the JSON request body WAS the record.

* fixed the format of the object in `errors`:
  * attribute `title` should be a string
  * there is no attribute `validation_error`.

* renamed route `get "/:model/:id/version/:version"` to `get "/:model/:id/versions/:version"`

* fixed route `get "/:model/:id/versions"`:

  * versions were stored as an array of objects in `data.attributes`, while this has to be hash.

  * `data` is now an array objects with the following structure:
    * `data.id` is the identifier of the record (so the same for all these objects)
    * `data.type` is the type of the record (so the same for all these objects)
    * `data.attributes` contains two attributes: `_id` and `_version`.
    * `data.links.self` contains an url to the version itself.

    => As you can see, this is a shorted list of the versions of a record.

* route `get "/:model/:id"` now adds an url to the list of versions in `data.relationships.versions.links.related`, if the model supports it.

* routes `get "/:model/:id/versions/:version"` and `get "/:model/:id/versions"` return an error if versioning is not supported for the model.

* validation errors (regarding the librecat model) now result in errors with pointers to the specific json path.

Some open questions:

* routes `put "/model/:id"` and `patch "/model/:id"` ignore `data.attributes._id` and overwrite it with `data.id`. Right?

* for model "publication" one can also update or delete the attribute "file", leaving the database out of sync with the filesystem.


nicolasfranck:
All the tests in `t/www/` now fail, except for `t/www/api.t`. 

Reason: all these failing tests expect preexisting data in the index/database, which is normally added by `t/www/00_preload.t`. Unfortunately `t/www/api.t` resets this data. Shouldn't each test take care of its own data, instead of implicitly relying on externally set data?

vpeil:
Also, we need to add the API documentation (in a separate PR). I can work on it asap this one is merged.

PR #882, opened 2020-03-16
LibreCat::Model->add should return updated record as bag->add always …

nicolasfranck:
The current implementation of `LibreCat::Model->add` returns either `1` or `0`, depending on the result. Shouldn't that be the updated record, as returned by `$model->bag->add`?


phochste:
On this one I would like to get @nics feedback why the modelling was made as it is.

Issue #816, opened 2019-10-25
links in popup "Open data librecat" in preview page not working.


nicolasfranck:
Reproduce:

* set status of record any other than "public"
* click on link "view" in the index page: you'll end up on a "preview page".
* click on the link "Open Data LibreCat", and choose one of the formats.

Why it does not work:

* it assumes it just has to add the format to the url, but while `/record/:id.:format` works, `/librecat/record/preview/:id.:format` does not. Why do we even put these records in a separate route?

nicolasfranck:
@phochste 

both routes `/record/:id` and `/librecat/record/preview/:id` use the same view `publication/record.tt`, so maybe I can fix this as following:

* deprecate route `/librecat/record/preview/:id`, meaning that this url is redirected to `/record/:id`
* `/record/:id` now filters on both `id=$id` and `status=public`. If I remove that last filter, then I can check if the user is logged in, and return access-denied if not. Problem: the requirement for login is determined by `config/permissions.yml`.

PR #885, opened 2020-03-18
add export controller for preview

nicolasfranck:
See also https://github.com/LibreCat/LibreCat/issues/816


nicolasfranck:
With this the route `/librecat/record/preview/:id` does not match anymore with routes like `/librecat/record/preview/1.json` (where 1.json is interpreted as id).

Issue #772, opened 2019-06-06
Need a stricter check on `new_record` in publication  updates


phochste:
`new_record` should only be possible for records that don't exist in the database. In all other cases, someone is messing around with the HTML form.

nicolasfranck:
`new_record` is only part of the problem. Now both new records and existing records are added by calling `POST /librecat/record/update`. 

What should happen for now:

* remove any reference to `new_record`
  difference between old and new records is the existence of an `_id`
* `_id` should only be generated when added to the bag, not when you open a web form.
    in case of UUID it doesn't matter, but it can be problematic for id generators that are
    number based, and just take the highest number, and add one to it.
    the models already have this built in, so need to replicate this.
* create route for posting new records: `POST /librecat/record`
   this should remove `_id` when present
* create route for updating existing records: `PUT /librecat/record/:_id`
   this should check for a matching record, and merge data with it.

willing to implement this, if everyone agrees.

phochste:
@nicolasfranck I agree. But you need to be very carefull. When uploading content to a 'new' record , LibreCat needs to know in some way the relation of the content to the metadata record. In the current implementation this relation is the record identifier. But please suggest a solution

nicolasfranck:
> @nicolasfranck I agree. But you need to be very carefull. When uploading content to a 'new' record , LibreCat needs to know in some way the relation of the content to the metadata record. In the current implementation this relation is the record identifier. But please suggest a solution

Are you referring to the file upload? The route `/upload` generates a new record id in the temp-store for every file that is uploaded, so when you upload 2 files, you get 2 directories with  each one file in it. That is the `tempid` that is referred to in `LibreCat::App::Catalogue::Route::upload`. So there is no need for `_id` to preexist in the form for this to work. Right?

phochste:
@nicolasfranck yes please I was referring to this. It will probably work indeed as you said, but please also test the file uploads when you start working on this

nicolasfranck:
see https://github.com/LibreCat/LibreCat/pull/888

nicolasfranck:
@phochste you were right about the need to insert the `_id` in the form. I saw that this is not required by the uploading module, but by `LibreCat::App::Catalogue::Controller::File::handle_file`. So I did not change that.

PR #888, opened 2020-03-19
Fix issue 772

nicolasfranck:
What I've done:

* Split route `post "/librecat/record/update"` in two routes:
  * `post "/librecat/record"`:
    * supplied `_id` should not have a corresponding record in the database yet.
    * `status` is set to `private`. This used to live in template `backend/generator/master`.
  * `put "/librecat/record/:id"`:
    * supplied `_id`must have a corresponding record in the database
    * Rules about `finalSubmit` apply only here
    * versionconflict can only happen here, not during a record creation.

* added dancer `var` to supply extra information to the form template so it can make decision like:
  * the action of the form: `var form_action => ..`
  * the method of the form: `var form_method => ..`
  * new_record: `var new_record => 0|1`
 
   These variables are accessible in template using the keyword `vars`.
   The form method `PUT` is normally not possible, but thanks to the already existing middleware `Plack::Middleware::MethodOverride` one can add the parameter `x-tunneled-method`, and use `POST` in the actual form.

* Only use body parameters when updating/creating records.

* Removed unused rule `$params->{meta}->{template}` (https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App/Catalogue/Route/publication.pm#L299)

* replaced `sprintf` methods by localize with parameters. See also the changes in `config/locale.en.yml` and `config/locale.de.yml`! I could not provide the German translations, so I've put English in locale.de.yml..

* Updated route `post /change_type`

* Updated route `get /clone/:id`

### DIFF
--- a/config/hooks.yml
+++ b/config/hooks.yml
@@ -27,6 +27,11 @@ hooks:
     before_fixes: []
     after_fixes: []
 
+  # Code that runs when creating a publication
+  publication-create:
+    before_fixes: []
+    after_fixes: []
+
   # Code that runs at a 'Make public' action
   publication-publish:
     before_fixes: []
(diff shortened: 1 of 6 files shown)


nicolasfranck:
https://github.com/LibreCat/LibreCat/issues/772

PR #893, opened 2020-03-24
Message bag

### LINKED ISSUE #735: Move the message to reviewer to separate database
User communication shouldn't enter the record metadata but stored in a separate database. It makes it also possible to create a whole conversation with the user.

And, pref, also add this functionality to the API

vpeil:
Fixes #735 

## What's in this PR

- Message field ist removed from the forms, the publication schema, etc.
- Theres's a new bag `message`.
- In the backend you see a new link "Message" for each publication record on the list.
- Now, enter your message and/or see the messages concerning this record.

## Questions:

- Migration needed. Provide script or is documentation in the wiki enough?
- Do we need a hook `message`. E.g. you want to put `imported from crossref` during an import. We could still put the info into the message field, but then the hook gets called, pushes this message info into the message-db and deletes the message field from the record. Problem: Record-ID must be known.


nicolasfranck:
@vpeil and maybe use the Catmandu plugin `Datestamps` for the bag `message` that automatically adds `date_created` and `date_updated`. Of course if you want to resubmit an old message using `add`, the validator needs to accept these attributes.

vpeil:
> @vpeil and maybe use the Catmandu plugin `Datestamps` for the bag `message` that automatically adds `date_created` and `date_updated`. Of course if you want to resubmit an old message using `add`, the validator needs to accept these attributes.

I just copied from LibreCat::Audit :smile:  But there is no use case where you have to update a message. So date_created and date_updated would always be identical.

nicolasfranck:
@vpeil I noticed that ;-). But it would still be useful to know the date at which a message was sent, although this information is not present yet in the current database.

vpeil:
The date is saved:

Bildschirmfoto_2020-03-26_08-57-12

This is done by line https://github.com/LibreCat/LibreCat/blob/message_bag/lib/LibreCat/Message.pm#L64

nicolasfranck:
I see, but wouldn't it be better that all or most timestamps are handled in the same way? I know, in this case it would depend on the configuration (i.e. use of plugin  Datestamps), but that goes also for the publications were dates are also important.

vpeil:
Agreed, but then also for audit!? For the sake of consistency.

nicolasfranck:
Right, dammit, I did forgot that myself. But hard to undo now without breaking stuff.

vpeil:
If we want to process a message field (e.g. at imports) we need to add this logic in

```perl
around add => sub {...};
```

in `LibreCat::Model::Publication`.

phochste (changes requested):
This is a good start to have a better handling of messages!

Certainly a conversion is needed if this is released in the main code (because the data model has changed). There are some interface and security considerations that might need attention:

Interface:

- It is not clear when new messages arrive for a record. As an end users they go quite unnoticed and currently one needs to click all the message links in the record to view them. 
     - At least a counter next to message link is required. E.g. message(3)
     - It would be better to have an indication if a record has "new" messages
     - Maybe an extra tab is required  in the interface to show all record which have messages
- Currently it is possible to inject any kind of HTML in the message (which shouldn't be the case)

Security:

- Currently any user is allowed to read any message via the API, and any user is allowed to post messages for any type of record via the API. This need to be restricted to the users that are allowed to edit the record.

Coding:

- Indeed a hook would be welcome when a message gets posted to the system.

vpeil:
> This is a good start to have a better handling of messages!
> 
> Certainly a conversion is needed if this is released in the main code (because the data model has changed). There are some interface and security considerations that might need attention:
> 

Yes, I would add a conversion script if the other issues with this PR can be solved.

> Interface:
> 
> * It is not clear when new messages arrive for a record. As an end users they go quite unnoticed and currently one needs to click all the message links in the record to view them.
>   
>   * At least a counter next to message link is required. E.g. message(3)
>   * It would be better to have an indication if a record has "new" messages
>   * Maybe an extra tab is required  in the interface to show all record which have messages
>
This is a delicate question: In general we don't know to whom a message is sent. User to reviewer, reviewer to admin, and vice versa and so on. ... But I agree that we need some better UI here. Maybe an extra tab (as you suggested) and show all incoming messages from your last login date (which we currently do not store)

> * Currently it is possible to inject any kind of HTML in the message (which shouldn't be the case)
> 
Will fix this.

> Security:
> 
> * Currently any user is allowed to read any message via the API, and any user is allowed to post messages for any type of record via the API. This need to be restricted to the users that are allowed to edit the record.
> 
I will add this security check.

> Coding:
> 
> * Indeed a hook would be welcome when a message gets posted to the system.

vpeil:
Security and Coding issues done.

Still to do:
- UI
- migration script

vpeil:
After our discussion some days ago, I am closing this one: it's a really huge change in the system. And probably not worth it at this point. But I'll leave the branch there.

PR #894, opened 2020-03-25
Fix issue 860

nicolasfranck:
What I've done:

* introduced limited set of attributes for the jwt payload (which is validated now):
  * `action`: 
    * required: false
    * type: array of strings
    * each item: enumeration. Valid: `show`, `create`, `update`, `patch` and `delete`. Note that access to a version is handled by action `show`. 
    * array cannot be empty if given
    * custom attribute
  * `model`:
    * type: string
    * enum: (lowercased) name of a model. e.g. `publication`
    * required: false
    * custom attribute
  * `cql`:
    * type: string
    * cql query that is validated against the `search_bag` of the corresponding model
    * required: false, unless the attribute `model` is given
    * custom attribute
  * `exp`
    * type: integer
    * minimum: 0
    * expiration time, expressed as time since Epoch
    * official attribute!
  * `nbf`
    * type: integer
    * minimum: 0
    * "not before"
    * official attribute!

* an empty payload means no  restrictions

* generated payloads are stored to the table `tokens`. This way they can be revoked

* `LibreCat::Token->decode($token)`:
  * decrypts token
  * validates payload attribute `exp` if present
  * validates payload attribute `nbf` if present
  * checks if payload is present in table `tokens`
  * return `undef` when checks above fail.
  * further validation of the payload are done by the routes responsible (see `json_api_v1.pm`)

* added tests

Questions/TODO:

* where to put the jwt validation in the routes? 
  * after the check "model exists" and "record exists": in that case one can check if records exist. status "403" means there was a hit!
  * before the check "model exists" and "records exists": in that case one cannot know if the record that was denied actually exists. 
* `action` `index` to list records, maybe filtered by `cql`, but that requires an extra route for which I can add another PR.


nicolasfranck:
I've also noted that the use `Crypt::JWT` is restricted by `LibreCat::Token`. The argument `alg` is hardcoded to `HS512`. Let it be or not?

vpeil:
> I've also noted that the use `Crypt::JWT` is restricted by `LibreCat::Token`. The argument `alg` is hardcoded to `HS512`. Let it be or not?

I guess, we could make it configurable and set "HS512" to default.

nicolasfranck:
@phochste 

* the api already returns status 403 when you try to access records that you do not have access to. Not?

* for the action `index`, what should we support? cql? I would prefer not the use the query behaviour that is used in the routes like `/librecat` or `/record` because that behaviour is flawed (see my pull request https://github.com/LibreCat/LibreCat/pull/869). cql can break easily, but I think such a syntax is acceptable in an api. And there is already an attribute `cql` in the token available.

* How can one "recreate" and old token? You probably refer to tokens with an expiration date (attribute `exp`), that have reached the expiration date? So copy the old token, and set that attribute to a higher value?

phochste:
* For the action `index` a CQL is support seems the correct choice indeed

* With `recreate` a token I don't mean literary creating a verbatim copy. Imagine just this situation:
     * I have a token database with dozens of tokens that were issues to people 
     * Some person has a problem with a token (something doesn't work as expected)
     * I need then a command to recreate the same token privileges as the person
     * I type `bin/librecat token recreate TOKEN-ID` and get a  token with exactly the same settings as the one with `TOKEN-ID` in the database. This token I can use in Postman or whatever tool to test what is happening for that person.

nicolasfranck:
Done:

* added subcommand `bin/librecat token export` to export all payloads to YAML
* added subcommand `bin/librecat token decode $token` to decode a given token
* added subcommand `bin/librecat token reencode $payload_id` to reencode a stored payload to a token
* added payload attribute `iss` (issuer) and `sub` (subject) (see jwt documentation)
* added api route `/api/v1/:model`:
  * pre filtered by jwt.cql
  * filtered by query parameter cql if provided

nicolasfranck:
A lot has changed:

* added action `index` (a name also use by rails for lists) and accompanying routes in the api: `GET /api/v1/:model`
* renamed `LibreCat::Token` to `LibreCat::JWTPayload` as that covers the meaning better. We only store the payloads, not the tokens that are given to the end users.
* the payloads are now stored in table `jwt_payload` instead of `jwt`
* renamed the command from `token` to `jwt_payload`. The meaning of the subcommands have changed, and some are added:

  * `bin/librecat jwt_payload add <YAML-FILE>` : 
    * add a list of payloads to the database table. Each payload is validated against a schema. Note that old payloads (i.e. having an `_id`) cannot/should not be added or updated like this: tokens are self contained json containers, so updating anything here does not affect the token itself. You either delete the payload or recreate one, but never update the meaning of it.
    * returns the payload record in YAML format to the stdout

  * `bin/librecat jwt_payload get <PAYLOAD-ID>`: returns the payload record in YAML format to the stdout

  * `bin/librecat jwt_payload export`: export all payload records in YAML format to the stdout

  * `bin/librecat jwt_payload encode <PAYLOAD-ID>`: encode an existing payload to a JWT

  * `bin/librecat jwt_payload decode <JWT>`: decode a JWT back to its payload. Note that validation is performed, and so expired tokens, or tokens whose payload has been deleted from the table are considered "revoked". For those, you get "cannot decode ..". Vague, but that's security. BTW. You can use `LibreCat::JWTPayload->new()->decode( $jwt, validate => 0 )` in order to decode expired tokens.

nicolasfranck:
I have have absolutely no idea why this travis build fails in perl 5.30: https://travis-ci.org/github/LibreCat/LibreCat/jobs/688770857

All perl tests are failing in that one build, and all seems to have the same error that is not related to my current changes, which were replacing two string in a comment: `Can't mkdir`.

Anyone any idea? This seems like an environment problem

nicolasfranck:
Pfew, seems to be fixed by simply waiting ;-)

phochste:
Can you add a `librecat jwt_payload delete`?

PR #897, opened 2020-03-27
add missing jsonschema errors

nicolasfranck:
What I've done: added missing json schema errors

TODO: German translations

@vpeil can you copy key `validator` and everything under it from `config/locale.en.yml` into `config/locale.de.yml` en provide the German translation?


vpeil:
Jaaaah.

Issue #900, opened 2020-04-14
Standard add a file.format_identifier metadata field to every file.


phochste:
For preservation workflows can calculate an identification of the stored file format that is more granular than the content-type of a file (which is currently provided by the client and not by the librecat software). While we don't have any support for tools like jHove, Droid in LibreCat, the result of these calculation could be available in some installations.

vpeil:
close this one (already merged into dev)

PR #903, opened 2020-04-16
Adding a free form file.data section where any JSON can be stored #900

phochste:
This is a solution for #900. A new field `data` field is added to the `file` section where any type of JSON blob can be stored for local usage. For instance, local implementers can use this field to store preservation metadata for files, or virus check information , DROID identifiers, or anything that can be imaged. 

This `file.data` can't be edited in the web forms or overwritten by these forms. The `file.data` can enter the system for instance by importing YAML files, using the `librecat publication files` command, or using the REST API.

If needed local implementers can make (parts) of this `file.data` searchable by adding the appropriate `index_publication.fix` fixes.

### DIFF
--- a/config/forms.yml
+++ b/config/forms.yml
@@ -125,7 +125,8 @@ dropzone_fields:
         - checksum
         - date_created
         - date_updated
-    # all the descriptive fields below can be edited
+        - data
+    # all the descriptive fields below can be edited in web forms
     # all other fields not in these two lists will be deleted
     descriptive:
         - relation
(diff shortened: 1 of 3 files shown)


phochste:
I wouldn't recommend for now to use hooks for file uploads. Not until we have a way to handle uploads in an asynchronous manner. 

These JSON fields are for now meant for instance in use-cases where:
 - Files are added to LibreCat via command line tools
 - Files are processed (validated, checked, etc etc) after they have been uploaded to the system

In the future, yes ..we certainly want a hook into processing of files. But the current webinterface is not written for this use-case: it will wait endlessly until all the hooks have been processed..which can mean a long time depending on the hooks.

PR #907, opened 2020-06-08
LibreCat::Search: remove implicit use of facets

nicolasfranck:


lib/LibreCat/Search.pm
```diff
@@ -56,7 +56,7 @@ sub search {
         sru_sortkeys => $self->_sru_sort($p->{sort}) // '',
         limit        => $self->_set_limit($p->{limit}),
         start        => $p->{start} // 0,
-        aggs => librecat->config->{facets}->{$bag_name} // {},
+        aggs         => $p->{facets} // {},
```
vpeil, line 59: better use "// +{}" ?

nicolasfranck: There is no ambiguity, so it does not make any difference

PR #924, opened 2020-09-21
Fix radio buttons unchecking when sorting authors with drag&drop

petrakohorst:
Strange behavior in the edit forms when sorting authors by drag&drop:

Dragging an author from the bottom of the list to the top unchecks the radio buttons (extern/internal author) in every other line below the dropped author.

Dragging other authors and mixing it up with a new field unchecks several other radio buttons, even though the attribute checked="checked" remains in the source code.

There seems to be a difference in jQuery between an attribute and a property of a radio button. I did not fully understand, how to use them correctly, but my change at least fixes the problem.

(https://stackoverflow.com/questions/15844852/is-there-a-bug-with-radio-buttons-in-jquery-1-9-1)

### DIFF
--- a/public/javascripts/librecat.js
+++ b/public/javascripts/librecat.js
@@ -566,6 +566,10 @@ $(function () {
                                 var newname = $(this).attr('name').replace(/\d+/g,index);
                                 $(this).attr('name', newname);
                         }
+                        /* when we change "name", radiobuttons/checkbuttons loose their flag
+                           but the old value is still traceable..
+                           this is just a temporary fix
+                        */
                         if($(this).attr('checked') || $(this).prop('checked')){
                               $(this).removeProp('checked');
                               $(this).removeAttr('checked');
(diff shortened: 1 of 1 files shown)


nicolasfranck:
@petrakohorst It has nothing to do with `.prop` or `.attr` or jquery in general.

If you comment/remove the lines that change attribute [name](https://github.com/LibreCat/LibreCat/blob/master/public/javascripts/librecat.js#L565) then
the problem goes away. But you need this, right?

I think the problem is this: a browser stores form state somewhere, by associating names with values.
If you change the values, then you are actually swapping their values, even if the html form says
otherwise. But then again: it does not change the other input values (not?).

cf. https://stackoverflow.com/questions/9898954/prevent-radio-buttons-from-losing-checked-state-when-updating-name-attribute-via

petrakohorst:
That is an interesting hint, I have checked it out...

If we remove the name attribute from the checkboxes altogether (it is not saved or needed anywhere else anyway), the problem disappears. But then the checkboxes don't work right anymore.

And if we leave only the name attributes unchanged (similar to what you suggested) when dragging and dropping, that will break the "add_field" mechanism. This mechanism relies on the correct order of indices in names and ids.

So I don't see a way to solve this problem with the help of the name attribute. But since my changes in this PR work (at least for our system ... does it work for you?), I'd suggest that we use this solution.   
What do you think?

PR #926, opened 2020-09-22
Change status of a 'new' record to 'private' when editing

petrakohorst:
This used to be the case but the behavior was changed in a recent commit https://github.com/LibreCat/LibreCat/commit/7b260b4d1c85d5931e310f0b24398355c47057d1

Now editing a 'new' record will leave the status 'new' unless it is made public.

What was the reason for this change? Or is there maybe even a functionality in the backend logic that should still change the status to 'private' on editing, but doesn't do this correctly?

### DIFF
--- a/views/backend/generator/master.tt
+++ b/views/backend/generator/master.tt
@@ -32,7 +32,7 @@
     <input type="hidden" name="type" value="[% type %]" id="id_type"/>
     <input type="hidden" name="_id" value="[% _id %]"/>
     <input type="hidden" name="_version" value="[% _version %]"/>
-    <input type="hidden" name="status" value="[% status %]"/>
+    <input type="hidden" name="status" value="[% IF !status OR status == "new" %]private[% ELSE %][% status %][% END %]"/>
     <input type="hidden" name="date_created" value="[% date_created %]" />
     <input type="hidden" name="creator.login" value="[% creator ? creator.login : thisPerson.login %]" />
     <input type="hidden" name="creator.id" value="[% creator ? creator.id : thisPerson._id %]" />


nicolasfranck:
@petrakohorst I probably was confused about that line: "if status is new, then it should be private.."
I guess "new" isn't supposed to be a real status and it was just there to mark a new record, right? You cannot set the status anywhere to "new".

petrakohorst:
@nicolasfranck "new" is a regular status, just not one a user can actively set. It is set for Quick and Easy publications, for our Bielefeld dissertations, for WoS imported records and for automatically imported records via ORCID sync. Basically any record that is (semi)automatically imported into the system and needs to be seen by the admin team insead of just vanishing in the masses of publications should have status "new".

But as soon as someone edits this record it should turn to a private one, because then it is no longer "new"/"unseen".   
You can filter for "new" records and our support team uses this to find records that need to be edited.

Whether this should be a status or rather a flag is a matter of opinion/discussion. One could argue that a status should be something a user can actively set for a record - just as all the other statuses. In the end it doesn't matter, how we mark new records, as long as we're able to.

nicolasfranck (approved):

nicolasfranck:
@phochste @nics what your opinion on this subject? Should this statement "set status from new to private after edit" apply in general, or is this just a choice? Yes, one can set this by adding a hook, but then again, there is no button to change the status value from new to private.

phochste:
@nicolasfranck  this has no direct effect for my GREP/Librecat project. 

In general, it is a valid reason to flag imported records (without any human edit) as 'new'. This is comparable with mailboxes : unread emails are also in a state comparable to new, as soon as you open it it gets into a state read. 

I don't need a direct need for a user to switch this state manually with a button, if it is implemented it should be automatic.

nics:
Seems sensible to me

nicolasfranck:
ok, I'll merge it then

Issue #711, opened 2019-03-08
Roles Delegate and Data Manager get mixed up


petrakohorst:
I stumbled upon a strange bug (tested on the demo server with user ```testdelegate```):

User has delegation rights for another person and also data manager rights for a department. These are two separate roles with separate rights attached to them.

As a delegate the user will see the correct publication list, but only those publications will be editable that are type ```research_data``` and have the department of the ```data_manager``` role. All other records will return ```access denied (403)``` when opened for editing.

You can currently check this problem out by logging into the demo server as user ```testdelegate``` and switching to the delegate role. This user is a delegate for ```Albert Einstein``` and has ```data_manager``` rights for the ```Department of Physics```.   
In Einstein's list of 175 publications, which schould all be editable, you can find only one that actually is. That's the only ```research_data``` publication, that happens to have the department ```Department of Physics```.

So those two roles get mixed up. Does anyone have an idea why?

petrakohorst:
A year and a half later I have actually stumbled upon the solution for this ticket!

Explanation:   
The sub ```by_user_role``` in ```/LibreCat/Access.pm``` checks for each role (```reviewer```, ```project_reviewer```, ```data_manager``` and ```delegate``` - in that order) if a user with this role has rights to edit/return/publish... the current publication and returns 1 if that is the case for any of these roles.

If a user has both ```data_manager``` AND ```delegate``` rights, the loop ```if $role eq "data_manager"``` is executed before the ```if $role eq "delegate"```.
But the first thing this "data_manager" loop checks is if the publication is ```research_data```. And if not, it returns 0 and prevents the "delegate" rights loop further down from ever being evaluated.

So, a ```delegate``` who happens to also have ```data_manager``` rights is restricted to only editing ```research_data``` by this line.7

The fix is simple:   
Instead of returning 0 unless type eq research_data, let the loop run and only return 1 if the user has data_manager rights for this publication AND the publication is research_data. I fixed this in one of the commits in PR #927 .
(While I was at it, I included scientific software in the data_manager role.)

---

BUT: this is also closely related to the second commit in PR #927 , which adresses the fact that all of those loops are executed no matter which role is actually in use at the time.

Issue #908, opened 2020-06-08
LibreCat::Model::User::find too complex


nicolasfranck:
The methode `find` in `LibreCat::Model::User` executes too many possible queries:

* two possible queries against ElasticSearch (search on id, and if that fails, search on login).
* a `get` against the datastore itself
* a `find_by_username`, of which the number of search requests is determined by the configuration.

This method is only used by `LibreCat::App::Helper::Helpers::get_person`, and only `LibreCat::App::Catalogue::Route::file` uses any other argument than the user_id: email.

What to do with this?

nicolasfranck:
This method is used by `LibreCat::App::Helper::Helpers::get_person`. I've tried to replace all instances of `get_person` by `main_user->get` where I was sure that the real user identifier was given: https://github.com/LibreCat/LibreCat/pull/910

nicolasfranck:
crap. Just realised that not all users are stored in the main user table. Sigh.

vpeil:
> crap. Just realised that not all users are stored in the main user table. Sigh.

Do you mean the case we discussed in #748 ?

nicolasfranck:
@vpeil yes. That ticket also explains that we have users that we have no control over, as we cannot change them.

Issue #748, opened 2019-05-23
user whitelisting not consistent


nicolasfranck:
The way whitelisting of users is done in librecat is not consistent: it makes promises that it cannot fulfil.

In the configuration you can have

```
user:
  sources:
    - store: store1
      bag: bag1
      username_attr: login
    - store: store2
      bag: bag2
      username_attr: orcid
```

This way you can:

* match "login" with attribute "login" from bag1
* match "login" with attribute "orcid" from bag2

That's all fine,
but is does not force you to look in the main user database.
The user that is returned by methods LibreCat::Model::User::find_by_username
and LibreCat::Model::User::get is not necessarily contained in the
internal user database.

Proposal:
  * deprecate current use of configuration option "user.sources":
  * force "store" to "search" or "main". Default is "main".
  * force "bag" to "user", meaning you ignore the configuration option

Problems:
  * what to do with the store "builtin_users"?

vpeil:
I understand the store builtin_users as for testing or development. This can be done differently, though. I can't imagine someone uses it in production

nics:
We use it in production to store the superuser

nics:
But @nicolasfranck is right, the feature is problematic, not one of my best ideas ;-)

phochste:
@nicolasfranck what is the relation of this ticket to https://github.com/LibreCat/LibreCat/pull/747 ? Does this PR close this issue , is it unrelated to this issue , or something else?

nicolasfranck:
@phochste issue #747 is about blocking the user that was selected. This is about how we should select a user.

nicolasfranck:
The method `get` and `add` of this model are also contradicting each other: while `add` updates data in the underlying bag and search_bag, the `get` retrieves data from the first bag in a list.

For example, put this in `config/catmandu.local.yml` .. 

```
user:
  sources:
    - store: builtin_users
      username_attr: login
    - store: search
      bag: user
      username_attr: login
```

..and you will strange behaviour:

* user "einstein" (id 1234) appears in the list of users in `/librecat/admin/account`
* if you try to update that record using the form, the changes will appear in `/librecat/admin/account`, but if you return to the form nothing of the like will be visible.
* reason: the form uses `user->get` to retrieve, which should be correct, were it not that the `get` is overwritten in `LibreCat::Model::User`..

nicolasfranck:
[...]

PR #927, opened 2020-09-23
Delegate and Datamanager conflict

petrakohorst:
As mentioned in ticket #711 the sub ```by_user_role``` in ```/LibreCat/Access.pm``` does not consider the currently chosen user role, but only checks if the user account contains ANY role that has rights to edit this publication.

So as a reviewer for example a user would also be able to edit (e.g. from the frontdoor, where the edit link is displayed according to the result of can_edit) a publication which they only have delegate rights to.

This is not a security issue, but more of a "let's not mix up the roles" issue.

Solution:   
In ```/LibreCat/Permission.pm``` in the sub ```_can_do_action``` (which is the central sub for all the can_do_ subs), pass the current user role to the access sub ```by_user_role```.

And in ```/LibreCat/Access.pm``` in the sub ```by_user_role``` (in addition to some minor renamings) add the simple check "next unless current role eq permitted role".   
So in this sub we only accept roles that are 1. permitted for this publication, 2. roles/rights the user actually has and now 3. only the role the user has currently chosen.

### DIFF
--- a/lib/LibreCat/Access.pm
+++ b/lib/LibreCat/Access.pm
@@ -79,7 +79,8 @@ sub by_user_role {
         elsif ($urole eq 'data_manager') {
             for my $id (@{$user->{data_manager}}) {
                 for my $iid (@{$pub->{department} // []}) {
-                    return 1 if $id->{_id} eq $iid->{_id};
+                    return 1 if $id->{_id} eq $iid->{_id} &&
+                    (is_same($pub->{type},'research_data') || is_same($pub->{type}, 'software'));
                     for my $tree (@{$iid->{tree} // []}) {
                         return 1 if $id->{_id} eq $tree->{_id} &&
                         (is_same($pub->{type},'research_data') || is_same($pub->{type}, 'software'));
--- a/lib/LibreCat/Access.pm
+++ b/lib/LibreCat/Access.pm
@@ -80,10 +80,10 @@ sub by_user_role {
             for my $id (@{$user->{data_manager}}) {
                 for my $iid (@{$pub->{department} // []}) {
                     return 1 if $id->{_id} eq $iid->{_id} &&
-                    (is_same($pub->{type},'research_data') || is_same($pub->{type}, 'software'));
+                    ($pub->{type} eq 'research_data' || $pub->{type} eq 'software');
                     for my $tree (@{$iid->{tree} // []}) {
                         return 1 if $id->{_id} eq $tree->{_id} &&
-                        (is_same($pub->{type},'research_data') || is_same($pub->{type}, 'software'));
+                        ($pub->{type} eq 'research_data' || $pub->{type} eq 'software');
                     }
                 }
             }
--- a/lib/LibreCat/Access.pm
+++ b/lib/LibreCat/Access.pm
@@ -76,12 +76,12 @@ sub by_user_role {
             }
         }
         elsif ($role eq 'data_manager') {
-            return 0 unless is_same($pub->{type},'research_data');
             for my $id (@{$user->{data_manager}}) {
                 for my $iid (@{$pub->{department} // []}) {
                     return 1 if $id->{_id} eq $iid->{_id};
                     for my $tree (@{$iid->{tree} // []}) {
-                        return 1 if $id->{_id} eq $tree->{_id};
+                        return 1 if $id->{_id} eq $tree->{_id} &&
+                        (is_same($pub->{type},'research_data') || is_same($pub->{type}, 'software'));
                     }
                 }
             }
--- a/lib/LibreCat/Access.pm
+++ b/lib/LibreCat/Access.pm
@@ -46,7 +46,7 @@ sub by_user_id {
 #  then the department id of the publication should match
 #  the publication record
 sub by_user_role {
-    my ($self, $pub, $user) = @_;
+    my ($self, $pub, $user, $role) = @_;
 
     return 0 unless $pub && $user;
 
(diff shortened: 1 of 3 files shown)


nicolasfranck:
@petrakohorst what do you mean by "choose their role"? When they change their role by clicking on one of the options in the upper right corner (settings -> change role)?

What is actually the reason behind this behaviour? Why do we want to change a role?

I've seen problems with session.role going out of sync, especially after updating it
in the database (test it out and you'll see)..

petrakohorst:
@nicolasfranck Yes, I meant "change their role"... not "choose their role" (which they are not allowed to do themselves for security reasons, of course).

**"Why do we want to change a role?"**   
In our local system in Bielefeld there are several use cases for changing a role.

1. ```super_admins``` often change their roles to ```user```, ```delegate``` or ```reviewer``` to be able to reproduce errors the user has found. We even go as far as giving ourselves the exact roles and permissions a reviewer or delegate has (exact departments the reviewer has rights to) so we are able to see exactly what the reviewer is seeing.
2. A ```reviewer``` or ```delegate``` sometimes needs to change their role to ```user``` in order to find "lost" publications. They will sometimes enter a publication for a professor or for a subdepartment but forget to link the person or enter the corresponding department in the form. Then they save the publication and view the list again as reviewer/delegate, the publication is gone.   
The solution for them is to change their role to ```user``` (which displays all publications they are ```creator``` for), edit and correct the publication and then they can switch back and continue editing the delegate list or the reviewer list.
3. Some users may have both a ```reviewer``` role for a (sub)department AND ```delegate``` rights for one or more professors. Depending on what they want to work on, they need to change their role to switch between their "workspaces" so to speak.

------

I have made the changes proposed in this PR in our local LibreCat installation (in the core) in September 2020 and it has been running like this ever since without any problems. I have not noticed ```session.role``` going out of sync in our system so far.

nicolasfranck:
@petrakohorst so in fact you want to able to act as another user? 

I was thinking about that some time ago, but there is one problem:
LibreCat does not enforce the [whitelist](https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App.pm#L176) (during authentication) to be
the same as the internal bag `users` (`find_by_username` uses config `user.sources` which can have different stores),
so you cannot just use the `users` to present a selectbox.

see also: https://github.com/LibreCat/LibreCat/issues/908, https://github.com/LibreCat/LibreCat/issues/748

petrakohorst:
> so in fact you want to able to act as another user?

Yes, that is basically the functionality of the role ```delegate```. But it is a "act with the same rights as" rather than a "act AS" another user. Account and login boundaries will not be crossed.

The same goes for ```reviewers```. They act with a different set of edit rights than their basic user role permits. But they do not act as another user.

How does this PR connect to the `find_by_usernames` function and the whitelist/user authentication problems? And what do you mean by "use the `users` to present a selectbox" - what kind of selectbox do you refer to?

nicolasfranck:
@petrakohorst if I want to create a "login as" functionality (for super admins), I need a source database and present the users in a select box (or something similar).

vpeil:
> @petrakohorst so in fact you want to able to act as another user?
> 
> I was thinking about that some time ago, but there is one problem:
> LibreCat does not enforce the [whitelist](https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App.pm#L176) (during authentication) to be
> the same as the internal bag `users` (`find_by_username` uses config `user.sources` which can have different stores),
> so you cannot just use the `users` to present a selectbox.
> 
> see also: #908, #748

No, this is not about "impersonation". It is about the following use case: 1 user has different roles, and depending in what role one is currently logged you can do or see stuff or not. While logged in, the user can change his or her role.

petrakohorst:
@nicolasfranck I think there was a misunderstanding. This PR is not about creating a new functionality. The system already does everything we need it to do as is. There is no need for a "login-as" functionality.   
We have users (secretaries) needing to edit publications of their professors. For this we grant them edit rights to those exact publications (delegate role) - that is the main use case.

Now, if this secretary also has the job of editing all research_data publications of her department (data_manager role), then being able to choose between those two roles will cause an error: The secretary/delegate is only able to edit publications of their professor - that are ALSO research_data publications with a certain department attached to it.   
It's a pretty wild mixup of the rights of the two roles. See also Ticket #711

That's what I have fixed in this PR.

nicolasfranck:
@nics @phochste I am not really familiar with this kind of authorization. Can you comment on this?

phochste:
@nicolasfranck Hi , I am also not familiar with the delegate / datamanager workflow. But the best strategy for you would be in this case:

- Check with @vpeil and @petrakohorst if they agree on the intentions of the workflow
- Check if you can run the code
- Check with @petrakohorst if you can simulate a test and check that this test works
- If you can confirm the test is working and code is in good shape and @vpeil and @petrakohorst agree that what the delegate/datamanager workflow intends to do is correct, then you can merge. If further issues might popup, they will be new tickets.

nicolasfranck:
@petrakohorst @vpeil as I understand, this PR wants two things:

* fix situation where either role data_manager or delegate is evaluated, and when one of those two fail, it completely fails
* restrict access to the role that is chosen in the UI

I confer with the first option, but not directly with the second. Why do we even want to have an option that says "restrict myself to this role today"? I know it is already in the code, but shouldn't things like that be transparent? It is not uncommon that users have multiple roles, and that the underlying rights overlap

nicolasfranck:
@petrakohorst right I see your comment above about why you want to change roles..

nicolasfranck (approved):

PR #932, opened 2020-10-22
cloning records: delete PID urn and set status to new

vpeil:
- especially: avoid making records public by cloning


nicolasfranck:
@vpeil is it possible that some plugin automatically rewrites some lines for you (indentation for example)? Because I see lots and lots of unrelated changes in your code, and that should not happen). Maybe close this PR, and start a new one?

vpeil:
ok, will submit a new one.

PR #936, opened 2021-01-12
New and improved crossref DOI import.

petrakohorst:
Due to the crossref DOI import getting slower and slower, this PR introduces a new import using a different crossref API.

Crossref has changed their APIs to be faster and more reliable if services using the APIs identify themselves correctly, so they can be contacted in case of any problems resulting from their use of the API. Anyone refusing to identify themselves (as we have been doing - unknowingly) will only have access to a slow and often overloaded server pool running the APIs.

The new crossref import for LibreCat not only implements (at least the most basic) identification via an email address - and thus is waaay faster -, but also switches from REST to XML API, which provides far better and more detailed metadata for each record.

Important notes:
1. Make sure to set a **real**, **working** and somehow **identifying** email address in ```catmandu.local.yml -> admin_email```. You should be reachable there, it's only fair.
2. The ```fixes/lookup/cr_type_mapping.fix``` is kind of a wild guess on my account. I have not been able to find a reliable list of publication types in use at crossref so far. This mapping, however, works for the types coming into LibreCat from crossref.
3. The method of first determining the agency (crossref or datacite) for a DOI and then requesting the metadata is no longer usable. The API used for the agency check is the same slow REST API we have been using all along and it will basically kill a request through timeout before it has even happened. The new way to identify the agency is to request a DOI from crossref directly. The resulting XML (there will always be one) will either contain ```status="resolved"``` or ```status="unresolved"```. In the latter case we then request the DOI from datacite
4. There was quite some pushing and shoving involved in getting the XML into hash format and into the fixer. Crossref has some (to me) strange xml tags in there that needed to be replaced in order to be read correctly. These changes are done by regex in the ```fetch``` sub in ```lib/LibreCat/FetchRecord/crossref.pm```. Any suggestions for a better/nicer/faster/more beautiful way of dealing with this XML are welcome!


petrakohorst:
Examples of info that's missing in the old API, but present in the new one:

In records of type book_chapter  ([old API](https://api.crossref.org/works/10.37626/GA9783959871785.0.02) vs [new API](https://www.crossref.org/openurl/?format=unixsd&pid=librecat@somewhere.org&id=doi:10.37626/GA9783959871785.0.02))
- editors of the book are missing
- ISBNs are missing
- page info is missing (at least in our importer, it is present in the metadata)
- sometimes volume and issue are missing

In records of type book ([old API](https://api.crossref.org/works/10.14361/9783839459058) vs [new API](https://www.crossref.org/openurl/?format=unixsd&pid=librecat@somewhere.org&id=doi:10.14361/9783839459058))
- series title is missing
- ISBNs and eISBNs are both imported as ISBNs without distinction
- language is missing (and displayed as default english)

In records of type journal_article:
- ISSNs and eISSNs are both imported as ISSN, while the new API distinguishes between ISSN and eISSN

The ISSN-thing could be fixed for the old API. It lists two fields ```"ISSN":["2747-5476","2749-1986"]``` and ```"issn-type":[{"value":"2747-5476","type":"print"},{"value":"2749-1986","type":"electronic"}]``` so we'd just need to evaluate the issn-type field.

There are probably more things that might even be different on single record level (not type-related). But there are also fields missing in the new API, e.g. the field "subject" that is mapped to keywords in the old import. This is missing from the new API completely.

The main point in using the new API, however, was that the old one was no longer working properly. At least in our system you needed 5-10 (or more) tries to get a result for a DOI at all.

nicolasfranck:
Made some changes:

* make the importer die when invalid content is returned (other than xml and status 200). No need to try and convert other input if the rest api cannot return a valid response.
* esthetical changes to your fix

About the code itself: I cannot really validate all possible inputs, so I trust you on this ;-) The only thing I'm worried about is missing mappings? Or a "reject" in that mapping that may return an `undef`

vpeil:
Not necessary to merge, since #985 has been merged.

PR #945, opened 2021-04-06
use full url in redirect; return after redirect

nicolasfranck:
* use full url in redirect. Not just for the sake of completeness, but also because the base url can contain a prefix, i.e. when served from behind a proxy
* return after a redirect. See this [line](https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App.pm#L193) for example: it still works for some reason (the redirect happens), even though the lines after it are also executed (put some log lines to prove it). I made it explicit to prevent future bugs.

BTW. I saw this [line](https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App/Search/Route/person.pm#L101) also:

```
redirect sprintf "%s%s", h->config->{person}->{staffdirectory},
            uri_escape($id);
```

Is is required to put a full base url in the config for this?

### DIFF
--- a/lib/LibreCat/App.pm
+++ b/lib/LibreCat/App.pm
@@ -190,7 +190,7 @@ Route the user will be sent to if login is required.
 get '/login' => sub {
 
     # what are you doing? you're already in.
-    redirect '/librecat' if session('user');
+    return redirect uri_for('/librecat') if session('user');
 
     # not logged in yet
     template 'login',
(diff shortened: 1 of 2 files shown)


petrakohorst:
The difference between `redirect $uri` and `return redirect $uri` puzzles me, too. Or rather it puzzles me why one would want to execute code after a redirect.

The staffdirectory redirect is a peculiar thing... where is this being used? Is it being used at all? It's a route, so users might link to it from the outside... if they knew it existed. But why link to a LibreCat URL that does nothing else but redirect you to the staff directory?

Taken as is, it would require a path or url in the config. You could put `/person` in the config and the route would add the person id and redirect to `/person/ID`. Or you can put an external (full) url in the config and the route redirects there plus the id.

PR #947, opened 2021-04-19
fix single search routes

nicolasfranck:
* a few routes in [search.pm](https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App/Catalogue/Route/search.pm) redirect to a more specific route without first testing if that request can be made:
  * `/reviewer -> /reviewer/:department_id`
  * `/project_reviewer -> /project_reviewer/:project_id`
  * `/data_manager -> /data_manager/:department_id`
  * `/delegate -> /delegate/:delegate_id`
* of course you'll have to visit these routes manually. In normal cases, you are not redirected from `/librecat` to `/reviewer/:department_id` without having a role of reviewer, or even having a department assigned to you. But paths should be written in a defensive way.
* `/delegate/:delegate_id` and `/data_manager/:department_id` did not check if the user had the right to visit that path. So technically one can access any delegate path, if one has at least one delegate.. Same goes for `/data_manager/:department_id`
* If you cannot access a specific path (e.g. `/project_reviewer/:project_id`) it should issue a "access denied" instead of returning to the first project you have access to (can this lead to an infinite redirect loop??).

### DIFF
--- a/lib/LibreCat/App/Catalogue/Route/search.pm
+++ b/lib/LibreCat/App/Catalogue/Route/search.pm
@@ -7,6 +7,7 @@ LibreCat::App::Catalog::Route::search
 =cut
 
 use Catmandu::Sane;
+use Catmandu::Util qw(:is);
 use Dancer qw/:syntax/;
 use LibreCat qw(searcher);
 use LibreCat::App::Helper;
@@ -84,7 +85,9 @@ Performs search for reviewer.
 =cut
 
     get '/reviewer' => sub {
-        my $account = h->get_person(session->{user});
+        my $user    = session("user") or forward("/access_denied");
+        my $account = h->get_person($user) or forward("/access_denied");
+        is_array_ref($account->{reviewer}) && scalar(@{ $account->{reviewer} }) or forward("/access_denied");
         redirect uri_for(
             "/librecat/search/reviewer/$account->{reviewer}->[0]->{_id}");
     };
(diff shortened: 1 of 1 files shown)


petrakohorst:
Is it possible that this branch does not yet have the latest changes from dev?

I tried to test delegate rights but got stuck on that old data_manager/delegate problem.

nicolasfranck:
> Is it possible that this branch does not yet have the latest changes from dev?
> 
> I tried to test delegate rights but got stuck on that old data_manager/delegate problem.

Uh good point ;-)

petrakohorst:
Okay, now it all works the way it should, as far as I can see.

You're making a good point about paths being written in a defensive way. An `access denied` still got issued - from somewhere - if you accessed a delegate or reviewer route without the right permissions. But now it's clearly set in each route.

----

However, I did find another bug concerning user roles that rely on departments. At the moment I am not able to edit publications as reviewer if the publication only contains a subdepartment of "my" reviewer department.

Since it doesn't directly relate to this PR, though, I will prepare another PR and provide a fix.

nicolasfranck:
> An `access denied` still got issued - from somewhere - if you accessed a delegate or reviewer route without the right permissions. But now it's clearly set in each route.
> 

Try the following:

* Add a delegate to a user
* Login as that user
* Go to `/librecat/search/delegate/$id` with a user `$id` that is not yours or in in your delegate list.
* In de dev branch you are allowed to this page (which shouldn't happen)

PR #949, opened 2021-04-23
Fix permissions bug in edit links (search result list) for reviewers.

petrakohorst:
**Problem:**   
As a reviewer for department `xyz` my list of "allowed" publications is displayed correctly. But some of the entries (if not all of them) are not editable. This happens when the entry does not have `xyz` itself as department, but some subdepartment of it.   
The reviewer role depends on the department tree to check if a publication belongs to department `xyz` and thus the reviewer has edit rights to it.

`_can_do_action` checks against publications from the main store, but this PR https://github.com/LibreCat/LibreCat/pull/804 moved the department tree from the main store to the index.

Quick fix:   
When checking for permissions in `_can_do_action`, don't check in main store, but in index. (Additionally: set "live" parameter in template.)

### DIFF
--- a/lib/LibreCat/App/Catalogue/Controller/Permission.pm
+++ b/lib/LibreCat/App/Catalogue/Controller/Permission.pm
@@ -63,7 +63,7 @@ sub _can_do_action {
 
     return 0 unless defined($user_id) && defined($role);
 
-    my $pub   = $opts->{live} ? h->main_publication->get($id) : get_cached_publication($id);
+    my $pub = $opts->{live} ? h->publication->get($id) : get_cached_publication($id);
 
     is_hash_ref($pub) or return 0;
 
--- a/lib/LibreCat/App/Catalogue/Controller/Permission.pm
+++ b/lib/LibreCat/App/Catalogue/Controller/Permission.pm
@@ -26,7 +26,7 @@ sub get_cached_publication {
 
     my $pub = cache()->get( "RECORD_${id}" );
     my $set_cache = !$pub;
-    $pub //= h->main_publication->get($id);
+    $pub //= h->publication->get($id);
 
     cache()->set( "RECORD_${id}", $pub) if $set_cache;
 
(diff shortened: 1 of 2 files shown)


nicolasfranck:
@petrakohorst what is the reason for setting the parameter to `live`?
The cache is expected to have short living objects, but if you refetch
every record from the index, after you fetched the same list, it is going
to be slow again..

Maybe I should remove the configuration option `expires_in` from
the configuration because this is internal stuff, and only a stupid "fix"
for not using the fetched list of publications in the first place.

petrakohorst:
The "live" parameter is used to avoid calling `get_cached_publication`, which retrieves the rec from the main store. 

Alternatively we might change `get_cached_publication` to retrieve the record from the index instead of the main store. Then we may be able to stick to calling cached publications here.

nicolasfranck:
> The "live" parameter is used to avoid calling `get_cached_publication`, which retrieves the rec from the main store.
> 
> Alternatively we might change `get_cached_publication` to retrieve the record from the index instead of the main store. Then we may be able to stick to calling cached publications here.

Well you just replaced `h->main_publication` by `h->publication` which is the search bag, so that would solve your problem, not?

The "live" parameter (added by @phochste ) is just some internal thing to avoid hitting the cache, and is primarily used when downloading a file.
Apart from that, I haven't seen a good reason for it, as the cache lifetime is actually low.

petrakohorst:
`my $pub = $opts->{live} ? h->publication->get($id) : get_cached_publication($id);` only uses the freshly changed `h->publication` if the parameter `live` is set.

Without `live = 1` it uses `get_cached_publication`, which fetches the record from the main store again.

nicolasfranck:
@petrakohorst the cache only lives for 5 seconds; it is only meant to speed up permission checking during the same request. I actually intend to add a dancer hook `after` that clears the cache on the end of the request (like Rails does for example).

petrakohorst:
So should I just change this line https://github.com/LibreCat/LibreCat/blob/master/lib/LibreCat/App/Catalogue/Controller/Permission.pm#L29 to `$pub //= h->publication->get($id);`?   
And then remove the live=1 changes from the template?

nicolasfranck:
@petrakohorst I prepared a [PR](https://github.com/LibreCat/LibreCat/pull/950) that would make the option `live` obsolete.

petrakohorst:
Perfect! With your latest changes to this PR everything works as it should!

PR #950, opened 2021-04-26
remove configuration option permissions.cache.expires_in

nicolasfranck:
Caching of records was only introduced to deal with
a problem in LibreCat: permissions are checked using
identifiers, and not the records themselves. So it was
a "fix" for this, but one should not rely on it, or configure
its behaviour. It should contain very short living objects
in memory.


nicolasfranck:
I've changed it: the cache is saved in a request `var`, which means that the cache is gone after the request
(that was what it was meant for in the first place). 
So option `live` would not be needed anymore @phochste.

nicolasfranck:
Apparently `CHI` serialises every entry using `freeze` and `thaw`. So if you request them a thousand times, that `thaw` happens a thousand times. Better to use a regular hash if it is short local memory only needed.

PR #951, opened 2021-04-26
new method "current_user"; also conditionally reloads session attributes

nicolasfranck:
Two part PR

* Adds new method `current_user` to `LibreCat::App::Helper::Helpers`

  * loads user record on every request and keeps it in memory for the rest of the request. Accessible through `h->current_user()`
  * Current user is fetched using `LibreCat::Model::User#find(session("user_id"))`. Normally I would have chosen `LibreCat::Model::User#bag->get(session("user_id"))` but that would not work if `session("user_id")` came from store `builtin_users`
  * I've replaced lines like `h->get_person(session("user_id"))` or `h->get_person(session("user"))` where applicable

* Keeps user attribute `date_updated` in session as `user_last_updated`. When this value is lower than `current_user()->{date_updated}`, it means that the session contains outdated information, and needs to be reloaded from the user record. This happens for example if an admin has reset someones roles/rights (and therefore "role" of "reviewer" is no longer right).


petrakohorst:
I see no problem with this PR.

I'm guessing it is optional to replace all instances of `h->get_person(session("user"))` in local layers? Or might something break if local layers still used the old call to the user account while the LibreCat core uses `current_user`?

vpeil (approved):

nicolasfranck:
@petrakohorst `h->get_person` is still a function, so older (overrides of) views still can call that function.

PR #960, opened 2021-06-04
respect other access conditions in mods export

vpeil:


nicolasfranck (approved):

vpeil:
> @vpeil the fix `fixes/to_mods.fix` does not specify a default value for `accessRestriction`, so I am not sure whether it is usefull to have a statement like this without content..

ah, right, will specify a default value.

PR #967, opened 2021-09-29
Fixing a security risk concerning the parameter 'style'

petrakohorst:
We were made aware that one of our query params ('style') can be abused for Cross-site scripting (XSS). In some places our code uses the parameter `qp.style` directly in the template. This opens a door to insert malicious code into the LibreCat webpage.

To test this bug on the demo server, mark a few publications and then call this link:

http://demo.librecat.org/marked?style=ama%22%3E%3Cimg%20src=a%20onerror=prompt(document.domain)%3E

A (harmless!) prompt will appear at the top of the page, that should not be there. All other parameters seem to be unaffected by this.

This PR removes all direct usage of `qp.style` and replaces it with a check against the allowed style names in config.


nicolasfranck:
@petrakohorst I guess, since https://github.com/LibreCat/LibreCat/pull/972, this one can be closed?

PR #972, opened 2022-06-13
h->current_style an h->default_style

nicolasfranck:
@petrakohorst @vpeil See https://github.com/LibreCat/LibreCat/pull/967

I've added some helper methods:

* `available_styles` (array reference of styles in sorted order)
* `current_style`: current_style as found in the URL. Validated as well. Returns `undef` when not found or invalid.
* `default_style`

I've used these helper methods in the controllers where now configuration checks are repeated.

I've used these helper methods in the views as `h.current_style` and `h.default_style` where applicable.


petrakohorst (approved):
This looks good! Tried it and found no problems.

PR #973, opened 2022-06-14
remove try/catch statements from _do_index and remove_alias

nicolasfranck:
Reason: client code should catch the errors, not the libraries


nicolasfranck:
@vpeil @petrakohorst  everything seems to work here; but maybe double check on your part

vpeil:
@nicolasfranck l've tested these changs, works fine for me.

PR #976, opened 2022-06-14
citation-block-link should be block link; no absolute position!

### LINKED ISSUE #940: citation-block-link covers title within citation-block
https://github.com/LibreCat/LibreCat/blob/master/views/hits.tt#L78 covers
the [first link](https://github.com/LibreCat/LibreCat/blob/master/views/hits.tt#L78) in a hit 
with a link without text. So no matter what you put in that first link, only the empty, javascript
controlled link will work.

See also https://github.com/LibreCat/LibreCat/blob/6652a5ef0d037852e334cfdaee97b769d0559cfd/scss/librecat.scss#L369.
There you can see that the empty link is absolutely positioned above the other link.

nicolasfranck:
Fixes #940 

DONE:

* `citation-block-link` is no longer an absolute positioned link (without content) that hovers above the rest (including the title). It is just a link with `display:block`. This way the `href` applies to the block, instead of only the text (so same effect, but cleaner). Had to regenerate the main.css. Keep in mind to clear your browser cache (which is another issue: the name of the main.css)
* removed onclick statements


petrakohorst:
I can't seem to reproduce the error with the link. Can you show screenshots of what happens? Or explain in more detail which div covers which link?

nicolasfranck:
The citation-block-div, which is a link with an empty body is hovering over the entire record block in the list view (because of width, height and position in css), including the title of the record, that also has a link (same link!), but which cannot be clicked. It is hard to see the difference though, but I removed the dependency on javascript (onclick). Now both the record link and the citation can be clicked separately. 

I just removes unneeded complicated code.

nicolasfranck:
I only seems to happen in routes like `/librecat/search/admin`.

Screenshot 2022-06-15 at 15 10 40

Issue #978, opened 2022-06-15
template insertion into javascript creates broken edit account page


nicolasfranck:
The edit account page seems to broken due to a "javascript error"

After some digging I found out that the function `selectButton` from https://github.com/LibreCat/LibreCat/blob/master/views/admin/generator/master_account.tt#L110 is the culprit.
There content from template https://github.com/LibreCat/LibreCat/blob/master/views/backend/generator/fields/subfields/department_select.tt
should be inserted, but it always returns an empty string with a newline (which is incorrect
javascript for some reason)

The function `selectButton` is there to make the "select" (left from an "+" button) work,
so you can select a department from a modal, instead of via the autocomplete, right?

The reason why the template returns empty data is probably the absence of
the helper method `search_department`. That has been removed?
Probably this is still present in the Bielefeld branch, where it originated, right?

Anyway, do we need two ways of selecting the same data, autocomplete and modal?
What should be done if the list of departments/projects is empty (not taken into account).

@vpeil @petrakohorst

petrakohorst:
The search_department helper method is indeed still present in the Bielefeld branch and in our edit forms we use the select modal for departments actively.

**For the LibreCat core I'd fully agree that it is not necessary to have two different methods for selecting a department. Most people will prefer to use the autocomplete anyway.**

The same problem exists for project_select.tt, by the way.

In general, the problem with the newline appears in our forms as well, even though our ```department_select.tt```does have content. For some reason there is an invisible newline in this template, which inserts a linebreak into a javascript "document.write()" command. And that causes the whole javascript to break.    
This newline cannot be removed by opening and saving the template... because there is no newline visible. Not even in vim.

The only solution for getting rid of this mystery-newline is to open the file ```department_select.tt```in vim and do the following:
```

:set binary
:set noeol
:wq
```

This removes the invisible newline from the template until it is opened in an editor again.

nicolasfranck:
@petrakohorst interesting. It is not the first time vim or other editors insert newlines after the last line;
so relying on the absence of newline is not safe. Pff not the use `[%-` (note the `-`) is helping. I'll look further
into it.

PR #979, opened 2022-06-20
trim inserted blocks

nicolasfranck:
Fixes newlines from issue #978


petrakohorst (approved):
This is a great solution! It works for the admin user form AND it fixes a problem we have been having with out js-embedded publication lists.

PR #983, opened 2022-07-25
do not insert user query into cql; q already added to q by h->extract_params

### LINKED ISSUE #981: Bug in search on person page
An interesting phenomenon has occurred on the person page in the LibreCat core. This bug can be recreated on the LibreCat Demo-Server.

When searching on a person page (```/person/ID```) with **more than one search term AND reaching more than one hit**, an error happens. The error message says "core - template - undef error - cql error: unknown first class relation".   
Unfortunately there is no more info to be found in logs or anywhere as to which template threw this error.

Example: James Dean's Person page on the Demo-Server: http://demo.librecat.org/person/3

- Search for "retrospective relative" and you will find one hit.
- Search for "assessing" and you will find two hits.

But:

- Search for "james dean" and you should find three hits - but you get an error
- Search for "assessing genetic" and you should find two hits - but you get an error
- Search for "retrospective space" and you should find two hits - but you get an error

It has nothing to do with the returned hits, it has nothing to do with the actual search terms.

I have tried to check templates for the error source (tried to remove several included templates or parts of templates), but to no avail.   
I removed the filters.tt and the mark_all_js.tt from being included in the other templates - but the errors are still there.

Any idea what might cause this?

nicolasfranck:
Fixes #981

### DIFF
--- a/lib/LibreCat/App/Helper.pm
+++ b/lib/LibreCat/App/Helper.pm
@@ -194,7 +194,6 @@ sub is_marked {
 sub all_marked {
     my ($self) = @_;
     my $p = $self->extract_params();
-    push @{$p->{cql}}, $p->{q} if $p->{q};
     push @{$p->{cql}}, "status=public";
 
     my $hits       = librecat->searcher->search('publication', $p);


petrakohorst (approved):
Works perfectly!

PR #987, opened 2022-10-21
strongify all authors when using 'show all' an frontdoor

vpeil:
a tiny change


vpeil:
sorry, that was something from our local layer. I'll close it.

